PR #2, opened 2022-11-15
Add support for pluggable components as a buildingblock criteria to become stable

mcandeia:
Signed-off-by: Marcos Candeia <marrcooos@gmail.com>

### DIFF
--- a/guides/api-design.md
+++ b/guides/api-design.md
@@ -140,3 +140,4 @@ For **building blocks** to progress, the following must also be true:
 * E2E tests must exercise _at least two different implementations_ of the building block's API
 * Conformance tests testing both positive and negative cases must be defined
 * Certification tests for multiple components implementing this API must be defined
+* When applicable (no hard constraints involved like performance or environment limitations), support for [pluggable components](https://docs.dapr.io/operations/components/pluggable-components/pluggable-components-overview/) - with or without sdks


guides/api-design.md
```diff
@@ -140,3 +140,4 @@ For **building blocks** to progress, the following must also be true:
 * E2E tests must exercise _at least two different implementations_ of the building block's API
 * Conformance tests testing both positive and negative cases must be defined
 * Certification tests for multiple components implementing this API must be defined
+* Support for non built-in components, either WASM or [pluggable components](https://docs.dapr.io/operations/components/pluggable-components/pluggable-components-overview/) with or without SDKs.
```
mukundansundar, line 143: is this a `must` or a recommendation? please rephrase this to follow that format.

mcandeia: Fixed @mukundansundar

guides/api-design.md
```diff
@@ -140,4 +140,4 @@ For **building blocks** to progress, the following must also be true:
 * E2E tests must exercise _at least two different implementations_ of the building block's API
 * Conformance tests testing both positive and negative cases must be defined
 * Certification tests for multiple components implementing this API must be defined
-* Support for non built-in components, either WASM or [pluggable components](https://docs.dapr.io/operations/components/pluggable-components/pluggable-components-overview/) with or without SDKs.
+* Must have support for non built-in components, either WASM or [pluggable components](https://docs.dapr.io/operations/components/pluggable-components/pluggable-components-overview/) with or without SDKs.
```
mukundansundar, line 143: @mcandeia So what all is involved in adding support for pluggable components?

mukundansundar: Should this be `should` instead of must? Not all APIs and building blocks might need to be supported via pluggable components.

mcandeia: Actually, the overall idea is that there should have a way to let users create Dapr components privately, without including in the default binary (built-in). Either using WASM or Pluggable Components.

Obviously, this is not true (currently) given the current state of the buildingblocks, but it should be in the future.

The idea is to include somehow in the process of promoting a buildingblock to stable, the viability for implementing pluggable components, otherwise it will have no guidance wether should allow for pluggable components or not.

mcandeia: cc: @artursouza

berndverst:
I support this.

PR #3, opened 2022-11-16
Add proposal for crypto building block

ItalyPaleAle:
Ported from dapr/dapr#4508


0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,644 @@
...
+    // Encrypt a small message and returns the ciphertext
+    Encrypt(
+        // Context that can be used to cancel the running operation
+        ctx context.Context, 
+        // Input plaintext
+        plaintext []byte,
+        // Encryption algorithm to use
+        algorithm string,
+        // Name (or name/version) of the key to use in the key vault
+        key string,
```
mukundansundar, line 214: should this be `keyName` instead ?

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,644 @@
...
+    // PEM (SPKI) (default)
+    PEM = 0;
+    // JSON (JSON Web Key) as string
+    JSON = 1;
+  }
+ 
+  // Name of the component
+  string component_name = 1 [json_name="component"];
+  // Name (or name/version) of the key to use in the key vault
+  string name = 2;
```
mukundansundar, line 413: should this be `keyName` instead?

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,644 @@
...
+// SubtleEncryptRequest is the request for SubtleEncrypt.
+message SubtleEncryptRequest {
+  // Name of the component
+  string component_name = 1 [json_name="component"];
+  // Message to encrypt.
+  bytes plaintext = 2;
+  // Algorithm to use, as in the JWA standard.
+  string algorithm = 3;
+  // Name (or name/version) of the key.
+  string key = 4;
```
mukundansundar, line 438: same as above

mukundansundar (commented):
Should subtle crypto ops be available via `dapr` CLI?

yaron2:
> > Should subtle crypto ops be available via dapr CLI?
> 
> I don't see why not. But keep in mind the high-level ops are probably going to be gRPC-only.

This might conflict with our criteria to make APIs stable, in guaranteeing they have both gRPC and HTTP implementations. @mukundansundar do we have that listed in API maturity criteria right now?

mukundansundar:
> > > Should subtle crypto ops be available via dapr CLI?
> > 
> > 
> > I don't see why not. But keep in mind the high-level ops are probably going to be gRPC-only.
> 
> This might conflict with our criteria to make APIs stable, in guaranteeing they have both gRPC and HTTP implementations. @mukundansundar do we have that listed in API maturity criteria right now?

@yaron2 

Currently in the API guidelines the requirements for a new API are for [proposals](https://github.com/dapr/proposals/blob/main/guides/api-design.md#proposal-requirements)

- Proposed contract for the API
  - HTTP and gRPC APIs **should** be consistent in behavior and user experience.

further below for [Alpha APIs](https://github.com/dapr/proposals/blob/main/guides/api-design.md#creation-of-new-apis)

- Both HTTP and gRPC protocols **should** be supported for the new API

Both of these are recommendations that they should be consistent and supported. Not a strong _requirement_.

ItalyPaleAle:
> This might conflict with our criteria to make APIs stable, in guaranteeing they have both gRPC and HTTP implementations. @mukundansundar do we have that listed in API maturity criteria right now?

**Note that subtle operations are supported for both and what I am writing below ONLY applies to the high-level encryption/decryption APIs**

The rationale for only supporting gRPC is in the proposal:

> These APIs will only be available over gRPC, because they will work over a stream of data from the client. An HTTP implementation would require keeping the entire message in-memory, which is not acceptable.

Imagine the situation where you're encrypting (or decrypting) a very large file (multiple MBs or GBs). If you were to do that via HTTP, you'd have to send your entire blob to the runtime, then wait for the runtime to process it, and then read the response back. Even in the best-case scenario, where the runtime flushes the data as soon as it's processed, you still have an amount of data equal to the length of the message in memory at some point.

Using gRPC we are able to use 2-way streaming. The Dapr SDK would chunk the data and send each chunk to the runtime. The runtime processes the chunk and sends it back (encrypted or decrypted). The runtime will never need to keep in memory more data than the size of a chunk (64KB in the proposal for the high-level encryption scheme). That's MUCH better than possibly holding GBs in memory. The Dapr Encryption Scheme proposed verifies the data chunk-by-chunk, so even during decryption, we are never flushing un-verified data to the caller and this is considered safe.

So my point on not supporting HTTP is essentially to protect users from themselves, as there will be the temptation of sending large files over HTTP and that will cause memory issues.

> This is not a problem for subtle APIs because they're not meant to be used to encrypt large chunks of data. In some cases it's not even possible (e.g. when using RSA you're limited to a handful of bytes).

PS: The guidelines use **should** to indicate that both HTTP and gRPC should be supported, but it's not a strict requirement. During discussion, we highlighted a bunch of scenarios where it would be not advisable to support both protocols, and I believe this is one of them (most scenarios involved the need for streaming)

berndverst:
I support this.

0002-RCBS-Crypto-building-block.md
ItalyPaleAle: This is just a name used in an internal API

olitomlinson:
I support the introduction of this building block

ItalyPaleAle:
@johnewart and I had a review with the "crypto board" at Microsoft to advise on the design of the high-level encryption scheme. Overall we got approval from them, albeit with some small tweaks. I have applied a few of them already, and will complete the work by EOW.

The crypto board did not comment on the APIs being exposed, as that's purely a Dapr decision, just the design of the encryption scheme, the keys and algorithms used, etc.

ItalyPaleAle:
Completed updating the document with feedback from the crypto board. The change was in the headers of the Dapr encryption scheme, which is now a JSON object that covers both the "plain-text" header and the "binary" header.

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,675 @@
...
+header || binary payload
+```
+
+### Header
+
+The **header** is human-readable and contains 3 items, each terminated by a line feed (0x0A) character:
+
+1. Name and version of the encryption scheme used. Currently, this is always `dapr.io/enc/v1`
+2. The manifest, which is a JSON object.
+3. The MAC for the header, base64-encoded
```
artursouza, line 112: So, it should be easy to add more to the metadata by appending and using the line feed as terminator, right?

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,675 @@
...
+	// ID of the cipher used.
+	// 0x01 = AES-GCM
+	// 0x02 = ChaCha20-Poly1305
+	Cipher int `json:"cph"`
+	// Random sequence of 7 bytes generated by a CSPRNG
+	NoncePrefix []byte `json:"np"`
+}
+```
+
+> Dapr will choose AES-GCM as cipher by default.
```
artursouza, line 145: These should probably be configured via the component's YAML, not? It might also vary per implementation since some components will not have all ciphers.

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,675 @@
...
+
+Where:
+
+- `nonce_prefix` (7 bytes) is the nonce prefix from the header
+- `i` (4 bytes) is the sequence number, as a 32-bit unsigned integer counter, encoded as big-endian. The first segment has sequence number 0, and it increases.
+- `last_segment` (1 byte) is `0x01` if this is the last segment, or `0x00` otherwise
+
+## Components
+
+Components in dapr/components-contrib implement low-level primitives only, while all higher-level operations are performed by the runtime, so they are executed in a consistent way across all backends/services. This is because the job of the components is limited to actually interacting with the key vaults, and everything else is best handled by the runtime.
```
artursouza, line 205: Agree 100%. So, there will be a `crypto` type for components and 2 APIs will consume that (subtle and high-level), correct?

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,675 @@
...
+- `POST /v1.0/subtlecrypto/[component]/encrypt` -> SubtleEncrypt
+- `POST /v1.0/subtlecrypto/[component]/decrypt` -> SubtleDecrypt
+- `POST /v1.0/subtlecrypto/[component]/wrapkey` -> SubtleWrapKey
+- `POST /v1.0/subtlecrypto/[component]/unwrapkey` -> SubtleUnwrapKey
+- `POST /v1.0/subtlecrypto/[component]/sign` -> SubtleSign
+- `POST /v1.0/subtlecrypto/[component]/verify` -> SubtleVerify
+
+> Note: URL will begin with `/v1.0-alpha1` while in preview
+
+**Currently, higher-level APIs are not offered via HTTP**. This is because it's not possible, using HTTP, to have a stream where Dapr sends back data to the client before the client's request is complete. Without that, we'd have to buffer the entire data in-memory in the runtime, which would be not practical when encrypting very large files.
```
artursouza, line 675: Same comment here. We should offer that for compatibility reasons even with the perf penalty. Make it clear that for higher performance, should use gRPC with the SDK preferably.

artursouza (dismissed):
LGTM overall. Just some feedback:

1. Give an example of a `crypto` component's YAML.
2. Add HTTP High-Level API too - even with the perf penalty. It can be handy for local development and debugging.
3. The high level API should be done first:
   1. It will "wow" most users and our target audience does not really want to deal with low level APIs. I would even argue that people using low level crypto APIs might not consider Dapr at all.
   2. It is a smaller exposed surface for us to quickly add to the SDKs.

0002-RCBS-Crypto-building-block.md
ItalyPaleAle: Additional metadata could be added to the manifest, which is a JSON object. However, changes to the metadata would require a v2 of the protocol.

Any metadata in particular you are concerned with?

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,675 @@
...
+- `unwrapkey`: un-wraps (decrypts) keys, performing the opposite of what `/wrap` does
+- `sign`: signs an arbitrary message using an asymmetric key stored in the vault (we could also consider offering HMAC here, using symmetric keys, although not widely supported by the vault services)
+- `verify`: verifies a digital signature over an arbitrary message, using an asymmetric key stored in the vault (same: we may be able to offer HMAC too)
+- `getkey`: this can be used only with asymmetric keys stored in the vault, and returns the public part of the key
+
+### High-level APIs
+
+The high-level APIs at launch will only include support for **encrypting** and **decrypting** messages (of arbitrary length), using the [Dapr encryption scheme](#dapr-encryption-scheme-daprioencv1).
+
+These APIs will only be available over gRPC, because they will work over a stream of data from the client. An HTTP implementation would require keeping the entire message in-memory, which is not acceptable (messages could be multiple GBs in size).
```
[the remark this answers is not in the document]
ItalyPaleAle: Ok. As discussed, we will implement for HTTP and make it clear that it must not be used for large payloads.

I think (I will double-check) this should be already limited by the maxBodySize property of Dapr.

0002-RCBS-Crypto-building-block.md
```diff
@@ -0,0 +1,675 @@
...
+- Keeping keys outside of applications. Applications never see key material, but can request the vault to perform operations with the keys.
+- Allowing greater separation of concerns. By using external vaults, only authorized teams can access private/shared key materials.
+- Simplify key management and key rotation. Keys are managed in the vault and outside of the application, and they can be rotated without needing the developers to be involved (or even without restarting the apps).
+- Enabling better audit logging to monitor when operations are performed with keys in the vault.
+
+### APIs: High-level vs subtle
+
+The building block features 2 kinds of operations:
+
+- Low-level or "subtle" (term frequently used to indicate low-level crypto operations; one example is how in browsers, low-level operations are in the [`crypto.subtle`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) object). These offer developers full control over the schemes that are used, allowing them to specify parameters, keys, modes of operations, etc.
```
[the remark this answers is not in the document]
ItalyPaleAle: Will get high-level APIs implemented at the same time

0002-RCBS-Crypto-building-block.md
ItalyPaleAle: Correct. The diagram shows the flow: https://github.com/dapr/proposals/blob/c8f398716815ab51b67478cadde579c8384c3e45/resources/0002-RCBS-Crypto-building-block/data-flow.png

artursouza:
One more point, we need a reference to the encryption key in the high-level API's payload, so old data can be decrypted after key rotation. Asking users to manage key rotation would make the high-level API not as useful.

0002-RCBS-Crypto-building-block.md
ItalyPaleAle: No, this is not per-component because this operation always happens in the Dapr sidecar and never in the key vault, for performance reasons.

We implement a sort of "envelope encryption" scheme. The vault is only used to wrap the file key. The actual encryption/decryption of the data happens with the file key within the Dapr sidecar, so we have full control over the algorithms that are offered, and it's independent from the component.

yaron2:
> The same key could have different names when stored locally and when in Azure Storage

Why the reference to Azure Storage specifically?

ItalyPaleAle:
> Why the reference to Azure Storage specifically?

Sorry, that was a lapse. I meant Azure Key Vault, and fixed my message now

ItalyPaleAle:
@artursouza I thought more about your ask to include the key ID into the manifest, and how I could make that work, and I am still not convinced.

I could find 3 problems:

1. "Hardcoding" the key name makes the solution not portable. Users would not be allowed to switch where the key is stored, for example migrating from a K8s secret to Azure Key Vault, or vice-versa.
2. Especially when using asymmetric ciphers, it should be totally acceptable for the keys to be stored in different places. For example, app A only encrypts messages, so it only needs the public part of the key: that can be stored in a JWKS document retrieved from a URL. While app B which can decrypt messages needs the private part, which may be stored in a vault.
3. The key's name in vaults is not necessarily stable. For example, in Azure Key Vault each key can have multiple "versions", and each version could have the same key or a different one. We support this in Dapr Crypto, by allowing specifying the key version after the name, for example `mykey/version1` (versions in AKV are actually SHA hashes, so not something the user can type). If you don't specify a version, it defaults to `latest`. This causes two issues:
   - If the latest version is "version1": `mykey`, `mykey/latest` and `mykey/version1` all map to the same key.
   - `mykey/latest` and `mykey` are possibly moving targets: they do not necessarily identify one specific key.
   - We can implement a "resolver" that converts "latest" to a specific version, but that requires performing an extra network call which is very slow.

Thoughts? If not, are we ok to get this proposal officially approved?

artursouza:
The only thing is to use the new names instead from the thread we had.

ItalyPaleAle:
The change is in place:

- decryptionKey
- omitDecryptionKeyName

ItalyPaleAle:
FYI I have made a [small change](https://github.com/dapr/proposals/pull/3/commits/34da83ff1a94b6977adcc1c93739a2ce54564540) to the proposal to clarify the algorithms support for key wrapping in the high-level scheme.

1. Explained that AES-KW will require using 256-bit keys. This is because our file key (that will be wrapped) is 256-bit in length, and AES-KW thus requires using 256-bit wrapping keys too (this is per RFC 3394)
1. Removed AES-GCM-SIV which was listed as a "maybe", and will not be implemented in the near future. No vault supports it, it's not implemented in the "local crypto", and it was not recommended by the MS crypto board during the review.

mukundansundar:
+1 binding

artursouza (approved):

artursouza (approved):

PR #6, opened 2022-11-18
enhance `dapr run` to run multiple applications

mukundansundar:
Signed-off-by: Mukundan Sundararajan <65565396+mukundansundar@users.noreply.github.com>


mukundansundar:
## Open Question

Should `dapr compose` be renamed to something else? 
Say `project`? 

So the commands become `dapr project up`. We will have a `project` configuration file defaulting to `dapr.yaml` name.

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+The DAPR_HTTP_PORT and DAPR_GRPC_PORT will be passed in as extra environment variables to the application that is being run. Those flags for `daprd` can be overridden in the config file above but that is optional as the `compose` command will randomly assign a free port for each of those ports, similar to how `run` command currently behaves.
+
+### Preedence rules
+
+For `env` field:
+> Note: In addition to the defined env fields the app also gets the `DAPR_HTTP_PORT` and `DAPR_GRPC_PORT` fields.
+
+- If no field is present, the environment variables of the current shell which executes the `compose` command is passed to the app and dir.
+- if `env` field is present in the `common` section, in addition to the shell environment variables, the `env` map defined will be passed to all `apps` and `daprd` sidecars.
+- if `env` field is present only in a particular `apps` section, any shell environment variables, `env` maps from `common` section and the `env` map for the current app will be passed to both the `app` and `daprd`.
```
akhilac1, line 152: Since the different apps will be triggered from the the same shell that the daemon is running, the env variables might conflict?

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Proposed command format
+
+Given the configuration file defined above, there should be a way to use the configuration file and run the differnt commands and `daprd` with the configuration given in the file.
+
+For this there will be a `compose` command that will be defined in the `dapr` CLI.
+
+### Constraints
+
+`compose` should mimic what `run` does for the most part. The defaults should not be different from `run` command with any new configuration options having sane documented least permissions based defaults.
```
akhilac1, line 171: Assuming default values will be provided for attributes like DAPR_HTTP_PORT, what is the mechanism for the user to identify this port, so that they could execute a HTTP request etc.? It would be good to document user experience for interacting with daprd launched with compose

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+When multiple applications need to be run, monitored for crash(error exit) and restarted, state of the applications and associated other processes need to be maintained.
+
+#### How `dapr` currently maintains state across commands
+
+For getting the list of `daprd` processes and apps and then being able to stop them, `dapr` CLI is currently making use of the the metadata API that `daprd` provides directly.
+
+For example, on executing `dapr run`, there will always be an associated `daprd` along with any application that is run for that command. The PID for the CLI process, and the app command associated with it are put as extended metadata in the associated `daprd` process.
+
+When `dapr list` is run, it iterates through the OS processes list, gets all the listed `daprd` processes, queries their metadata API for getting details on the associated CLI PID, and APP command.
```
akhilac1, line 184: ps -aef | grep <pid> with pid as the process id of the daemon, gives the list of processes whose parent is the daemon process. This would be a good start but would require to remember the pid of the daemon process. Can this be returned as a token or a id for every start?
P.S : This is very specific to Linux and I am hoping there should be an equivalent for each OS

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+<checkout> ==APP== processing request 2
+<order-proc> ==APP== processing request 3. request 3 calls dapr API
+<order-proc> INFO[0000] requst 3 is calls dapr API             app_id=order-proc instance=Mukundans-MacBook-Pro.local scope=dapr.runtime.grpc.api type=log ver=1.9.3
+<checkout> ==APP== processing request 3
+
+```
+
+The logs will be as shown above, it will be chaotic to see what application is in which state and so on.
+
+Instead of writing to STDOUT, each application and associated `daprd` process will write the logs to `{application dir}/.dapr/logs/{app id}/app_{datetime}.log`, `{application dir}/.dapr/logs/{app id}/daprd_{datetime}.log`.
```
akhilac1, line 237: Is the user expected to look at the latest log to get the key attributes like DAPR HTTP PORT? This does not seem to be a good experience. Can you please check if key attributes required for the user to interact with Dapr and available as output of the dapr compose command?

akhilac1 (changes requested):

nyemade-uversky:
@dapr/maintainers-cli Putting this on your radar for 1.10

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+`dapr compose` will be a separate command in `dapr` CLI and is expected to coexist with `dapr run`.
+
+Compatibility with `dapr run` is expected to be maintained. But in certain cases there might be introduction of new behavior which might be _opt-in_ for `dapr run` whereas it might be _on by default_ for `dapr compose`.
+
+The expectation is for the behavior of `dapr compose` to be refined and stabilized over a series of releases starting from initial release in **v1.10**.
+
+### Recommendation for 1.10
+
+- Initial implementation will only support Linux OS.
```
nyemade-uversky, line 271:
```suggestion
- Initial implementation will only support Linux OS.
```
This should support both MacOS and Linux OS

nyemade-uversky (changes requested):

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+The scope of this proposal is to define a `compose` CLI command which will allow users to define and run multiple services from a single `compose` configuration file.
+
+This proposal specifically targets running in local environments and `slim` mode, where container engines are available. For running `daprd` container along with `app` container the solution is to use Kubernetes or docker-compose.
+
+For this proposal we will targetting runnning the applications and side cars as processes in the OS.
+
+Pros/Cons of using processes:
+- Pro: All other commands in `dapr` CLI for self hosted mode are written to work with processes
+- Con: Process execution is OS specific and it does not provide any isolation guarantees
```
yaron2, line 58: I don't think isolation should be called out as a con - process level isolation is outside the scope of Dapr and even containers are regarded as non-secure, semi-isolated mechanisms.

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+
+## Requirements
+
+The main requirements for the command:
+- being able to configure multiple dapr apps from a single configuration file
+- once the command is run with designated configuration file, if any application crashes, it will be restarted with an exponential backoff
```
yaron2, line 65: > if any application crashes, it will be restarted with an exponential backoff

I'm worried we will be creating and maintaining a process orchestrator here when the main goal of this is to provide a better developer experience to get started.

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+  - `app_id` application ID (mandatory field). Passed to `daprd` as `--app-id`.
+  - `app_dir` directory of the application (mandatory field).
+  - `resources_dir` (optional) directory(ies) of all dapr resources (components, resiliency policies, subscription crds) (overrides common def). Passed to `daprd` as `--resources-dir`.
+  - `config_file` (optional) the configuration file to be used for this app (overrides common def). Passed to `daprd` as `--config-file`.
+  - `app_protocol` Application protocol, HTTP, gRPC defaults to HTTP. Passed to `daprd` as `--app-protocol`.
+  - `app_port` port the app listens to if any. Passed to `daprd` as `--app-port`.
+  - other dapr run parameters (mostly pass through flags to `daprd`) All properties must have `_` as separators which will be validated(so that no unknown flags are passed) and translated to `-` for cmd line arguments for `daprd`.
+  - `command` ["exec" "arg1" "arg2"] format for application command
+  - `env` which overrides or adds to common env var defined or the shell env var passed in when `dapr compose` is called
+  - `restart` field which lets `dapr compose` know if the app/sidecar should be restarted on crash or not
```
yaron2, line 141: If we allow for automatic restarts, why would a user choose to not have it if they can?

0003-I-dapr-compose.md
yaron2: We should be mounting these env vars on the user process.

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+The execution of the above command uses the `<path/to/file.yaml>` as the `compose` configuration file or errors out if that file is not found.
+
+
+## Feature lifecycle outline
+
+`dapr compose` will be a separate command in `dapr` CLI and is expected to coexist with `dapr run`.
+
+Compatibility with `dapr run` is expected to be maintained. But in certain cases there might be introduction of new behavior which might be _opt-in_ for `dapr run` whereas it might be _on by default_ for `dapr compose`.
+
+The expectation is for the behavior of `dapr compose` to be refined and stabilized over a series of releases starting from initial release in **v1.10**.
```
yaron2, line 267: I would advise to remove the expectations about the milestone for this feature and the actual feature proposal.

0003-I-dapr-compose.md
msfussell: Agreed. This proposal should play no part in restarting a failed application. It should be up to the developer to see what apps are running and be able to issue a run command to simply restart any failed apps. Otherwise, you are building a process monitor/orchestrator.

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+This increases the complexity of using dapr in development, where users want to run multiple services in local mode and be able to partially/fully replicate the production sceanrio.
+
+In K8s mode this is alleviated through the use of helm/deployment YAML files. There is currently no such capability available for local self hosted mode.
+
+Asking a user to run multiple different `dapr run` commands each with different flags, increases the complxity for users onboarding onto Dapr.
+
+### Why dapr CLI?
+
+From the initial [proposal](https://github.com/dapr/community/issues/207), the solution was proposed as a seprate repo and CLI in itself. But later it was suggested to use dapr CLI itself to have a `compose` command in it instead of a having a separate CLI.
```
msfussell, line 26: using the word `compose` has so many connotation to Docker Compose, it leads to think this is only about running container. Given this is about mostly running processes I would use a different word, such as `dapr start`

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+```
+
+The logs will be as shown above, it will be chaotic to see what application is in which state and so on.
+
+Instead of writing to STDOUT, each application and associated `daprd` process will write the logs to `{application dir}/.dapr/logs/{app id}/app_{datetime}.log`, `{application dir}/.dapr/logs/{app id}/daprd_{datetime}.log`.
+
+### Subcommands and flags for `compose` command in CLI
+
+It is proposed to have a subcommand `up` in `compose` command to _bring up_ all the applications and daprd sidecars. Later on an extension can be added to _shutdown all_ application and sidecars with a `down` subcommand.
```
msfussell, line 241: Why introduce the 'up' command and simply have just `compose` to start all the applications?

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+> Note: For application crash scenario, the only way in CLI process to discover that the application has crashed is through the process exiting with a non-zero status code.
+
+## Proposed Structure for organizing Dapr projects locally
+
+Currently `dapr` CLI initializes in the home directory(user profile dir for windows) a folder called `.dapr` and the default configurations and resources to be used are stored there.
+
+Users developing different apps using dapr will have different resources/configurations that they use per application. Each time the user has to run the application with a particular config and resources directory value, they have to override the flag for `dapr run` command.
+
+Instead the following convention is proposed for loading the resources/config for an application.
```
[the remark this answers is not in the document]
mukundansundar: Could you explain this part @akhilac1?

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+  - `app_protocol` Application protocol, HTTP, gRPC defaults to HTTP. Passed to `daprd` as `--app-protocol`.
+  - `app_port` port the app listens to if any. Passed to `daprd` as `--app-port`.
+  - other dapr run parameters (mostly pass through flags to `daprd`) All properties must have `_` as separators which will be validated(so that no unknown flags are passed) and translated to `-` for cmd line arguments for `daprd`.
+  - `command` ["exec" "arg1" "arg2"] format for application command
+  - `env` which overrides or adds to common env var defined or the shell env var passed in when `dapr compose` is called
+  - `restart` field which lets `dapr compose` know if the app/sidecar should be restarted on crash or not
+
+The DAPR_HTTP_PORT and DAPR_GRPC_PORT will be passed in as extra environment variables to the application that is being run. Those flags for `daprd` can be overridden in the config file above but that is optional as the `compose` command will randomly assign a free port for each of those ports, similar to how `run` command currently behaves.
```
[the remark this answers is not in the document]
mukundansundar: Generally Metrics port will not be consumed by the application directly. Metrics port though will be available as needed similar to how it is done in `dapr run`.

0003-I-dapr-compose.md
mukundansundar: Both DAPR_HTTP_PORT and DAPR_GRPC_PORT will be passed as env vars. I do not think metrics is port is needed inside the app, that can be part of the `dapr compose` output?

0003-I-dapr-compose.md
mukundansundar: We are not running `ps` command since as you mentioned it is Linux specific, but we do use a library which returns all running processes.

0003-I-dapr-compose.md
mukundansundar: Agreed ... That is what I was thinking as well.

0003-I-dapr-compose.md
mukundansundar: Agreed removed.

0003-I-dapr-compose.md
mukundansundar: I added this part, since it was mentioned as one of the main requirements in the initial community proposal itself. https://github.com/dapr/community/issues/207. First point mentions keeping track of termination and restart policies. 

I completely agree that if we go this route we will be going for a process orchestrator and the idea behind the _daemon monitor_ process mentioned as one of the options is that only. 

@artursouza thoughts?

0003-I-dapr-compose.md
```diff
@@ -0,0 +1,294 @@
...
+
+When `dapr stop` is called on an APP ID, it calls list and gets that particular APP ID based CLI PID/ daprd PID and kills that.
+
+#### Constraint with `compose`
+
+`compose` is also planned as a command which runs multiple applications, that will make it consistent with existing CLI commands.
+
+But one of the requirements for this feature is that there should be monitoring capability which restarts the application on a crash (process exited with an error).
+
+In this case, the ideal solution is to have a _daemon service_ that is independent of `dapr` CLI, which monitors any and all applications that are run when executing `dapr compose`.
```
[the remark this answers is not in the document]
mukundansundar: see above.

0003-I-dapr-compose.md
mukundansundar: the only concern that I have here is, we would have too many `start` related commands. 
`dapr init`, `dapr start`, `dapr run`. 

Also `dapr start` does in no way convey that it is a command to start multiple apps in a single go.

0003-I-dapr-compose.md
mukundansundar: This was mostly following the same way `docker compose`, it will be familiar to the users.

mukundansundar:
@yaron2 @artursouza @msfussell Based on the offline discussion that we had, we tentatively agreed to use `dapr run -f .` as the flag under which we will enable this feature and not under `dapr compose`. The main concern that I have with this is that, this is no longer an individual command and we will be modifying an existing command. Is that something that we want to go for? This proposal from start assumes that the command will be independent and will not affect any existing commands. 

Thoughts?

yaron2:
> 

I'm not concerned with changing the existing command, if the `-f` flag targets a completely different path. I think this is an implementation detail that shouldn't be discussed here - as far as the user experience goes, `dapr run -f` is the desired experience we'd like to see.

msfussell:
I do like the approach of extending the existing `dapr run` command, which was why I questioned the need to have the `up` command. I suggest writing out full examples for using `dapr run -f` and `dapr stop -f` or `dapr stop --all` as this then builds on the current CLI commands.

mukundansundar:
> I do like the approach of extending the existing `dapr run` command, which was why I questioned the need to have the `up` command. I suggest writing out full examples for using `dapr run -f` and `dapr stop -f` or `dapr stop --all` as this then builds on the current CLI commands.

To summarize, the proposal should be modified to
-  not have process orchestration
- extend `dapr run` to support running multiple files defined in an ingested `"file"`
- define the file structure for the `"file"`
- define a default directory struct for defining components to use with this feature
- output logs to a file instead of STDOUT. (Let users know of file location for logs)

@yaron2 @msfussell @artursouza 
Is this correct?

yaron2:
> > I do like the approach of extending the existing `dapr run` command, which was why I questioned the need to have the `up` command. I suggest writing out full examples for using `dapr run -f` and `dapr stop -f` or `dapr stop --all` as this then builds on the current CLI commands.
> 
> To summarize, the proposal should be modified to
> 
> * not have process orchestration
> * extend `dapr run` to support running multiple files defined in an ingested `"file"`
> * define the file structure for the `"file"`
> * define a default directory struct for defining components to use with this feature
> * output logs to a file instead of STDOUT. (Let users know of file location for logs)
> 
> @yaron2 @msfussell @artursouza Is this correct?

LGTM.

0003-I-dapr-compose.md
karolz-ms: `dapr up`?

yaron2:
+1 binding

mukundansundar:
+1 binding

yaron2 (approved):

msfussell:
> > I do like the approach of extending the existing `dapr run` command, which was why I questioned the need to have the `up` command. I suggest writing out full examples for using `dapr run -f` and `dapr stop -f` or `dapr stop --all` as this then builds on the current CLI commands.
> 
> To summarize, the proposal should be modified to
> 
> * not have process orchestration
> * extend `dapr run` to support running multiple files defined in an ingested `"file"`
> * define the file structure for the `"file"`
> * define a default directory struct for defining components to use with this feature
> * output logs to a file instead of STDOUT. (Let users know of file location for logs)
> 
> @yaron2 @msfussell @artursouza Is this correct?

Agreed. LGTM

PR #9, opened 2022-11-21
Pluggable components injector proposal

mcandeia:
Signed-off-by: Marcos Candeia <marrcooos@gmail.com>

### DIFF
--- a/0005-R-pluggable-components-injector.md
+++ b/0005-R-pluggable-components-injector.md
@@ -1,0 +1,298 @@
+# Pluggable components injector
+
+- Author(s): Marcos Candeia (@mcandeia)
+- State: Ready for Implementation
+- Updated: 11/21/2022
+
+## Overview
+
+Pluggable components are components that are not included as part of the runtime, as opposed to built-in ones that are included. The major difference between pluggable components and built-in components is the operational burden related to bootstrap/start the pluggable component process that are not necessary when using a built-in one since they run in the same process as Dapr runtime. This operational burden is present in many ways when using pluggable components and can lead to errors and hard debugging. In addition, there are certain configurations that are tied to the Dapr and how the runtime registers the pluggable component that is repetitive and can be better handled by Dapr instead of delegating this responsibility to the end-user. This proposal suggest the addition of a new mode of execution for selected pluggable components: injectable pluggable components.
+
+## Background
+
+#### Decrease the operational burden
+
+Even considering the new pluggable components annotation from [#5402](https://github.com/dapr/dapr/issues/5402), setting up applications to properly work with pluggable components still not an easy task due to the operational related to bootstrapping containers over and over again for each application that the user needs, especially if you consider that components are often not well [scoped](https://docs.dapr.io/operations/components/component-scopes/). Without scope, a component make itself available for all applications within the same namespace, meaning that every deployment/pod should re-do the same manual job of mounting volumes, declaring environment variables and pinning container images.
+
+So let's say you have an application named `my-app` and, another one named `my-app-2`, your two deployments/pods will look like the following:
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+name: app
+labels:
+  app: app
+spec:
+replicas: 1
+selector:
+  matchLabels:
+    app: app
+template:
+  metadata:
+    labels:
+      app: app
+    annotations:
+      dapr.io/pluggable-components: "component"
+      dapr.io/app-id: "my-app"
+      dapr.io/enabled: "true"
+  spec:
+    volumes:
+      - name: my-component-required-volume
+        emptyDir: {}
+    containers:
+      - name: my-app
+        image: my-app-image:latest
+      ### This is the pluggable component container.
+      - name: component
+        image: component:v1.0.0
+        volumes:
+          - name: my-component-required-volume
+            mountPath: "/my-data"
+        env:
+          - name: MY_ENV_VAR_NAME
+            value: MY_ENV_VAR_VALUE
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+name: app-2
+labels:
+  app: app-2
+spec:
+replicas: 1
+selector:
+  matchLabels:
+    app: app-2
+template:
+  metadata:
+    labels:
+      app: app-2
+    annotations:
+      dapr.io/pluggable-components: "component"
+      dapr.io/app-id: "my-app-2"
+      dapr.io/enabled: "true"
+  spec:
+    volumes:
+      - name: my-component-required-volume
+        emptyDir: {}
+    containers:
+      - name: my-app-2
+        image: my-app-2-image:latest
+      ### This is the pluggable component container.
+      - name: component
+        image: component:v1.0.0
+        volumes:
+          - name: my-component-required-volume
+            mountPath: "/my-data"
+        env:
+          - name: MY_ENV_VAR_NAME
+            value: MY_ENV_VAR_VALUE
+```
+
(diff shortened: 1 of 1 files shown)


mcandeia:
cc: @yaron2

0005-R-pluggable-components-injector.md
```diff
@@ -0,0 +1,298 @@
...
+
+And the component spec:
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: my-component
+  annotations:
+    dapr.io/component-container-image: "component:v1.0.0"
```
yaron2, line 265: How is this going to get injected into the pod? would the sidecar injector now need to get a list of all Dapr components in the namespace and create the corresponding containers in the pod?

0005-R-pluggable-components-injector.md
```diff
@@ -0,0 +1,298 @@
...
+replicas: 1
+selector:
+  matchLabels:
+    app: app
+template:
+  metadata:
+    labels:
+      app: app
+    annotations:
+      dapr.io/inject-pluggable-components: "true"
```
yaron2, line 224: I don't see how this solves the scoping problem. What happens if you have 2 pluggable component and two apps - each app needing a different pluggable component? This annotation would inject both pluggable components to a deployment.

0005-R-pluggable-components-injector.md
mcandeia: There are few alternatives,

Ideally, we should guarantee that whatever the component list will be fetched from, it should be consistent with the runtime component list to avoid racing conditions between the former and the latter, e.g in case of a pluggable component is configured between the injector patch and the pod start. 

Even considering that racing condition, the reconciling k8s philosophy will help to reach the desirable state somehow.

In the case of consistency being a `MUST HAVE` I see two phases to achieve that

phase 1: Rely on k8s

1. Use the daprClient to fetch the component list and apply the filter by scope logic 
* Advantage:
    - do not depend on operator to work
* Disadvantage:
    - have to "duplicate" the logic to filter by scope given the app name
2. Fetch from operator API
* Advantage:
    - do not have to "duplicate" the logic to filter by scope given the app name
* Disadvantage:
    - depend on operator to work

phase 2: Consistency guaranteed

I don't have a clear path here, it could be something like a `components version` that allow to fetch a specific version of the dapr configuration state, or something like we've discussed about having a init container that populates the components directory based on a base64-encoded components yaml.


----

What do you think @yaron2  ?

0005-R-pluggable-components-injector.md
mcandeia: > This annotation would inject both pluggable components to a deployment.
Yes, this annotation hasn't the purpose of solving scope, it is basically used to enabling/disabling the pluggable components injector, similar from what we have for the sidecar injector.

The "scope" problem that is solved by this proposal is the similarity with how the built-in components work today. You don't have to worry about scope them because it just works. Differently from what pluggable components offers, which kinda obligates the user to properly scope pluggable components due to the container's dependency on their pods.

ItalyPaleAle:
This may not be the best place for this conversation, but reading this... Wouldn't it make sense to look into creating a specific CRD for Dapr-ized apps instead? Something like `DaprApp`?

We had a conversation about this on Discord a while back: https://discord.com/channels/778680217417809931/1033131188128317520/1033131188128317520

It's not that I'm against this proposal, which (with some changes perhaps as Yaron is discussing) is probably the best we can do right now. But we are starting to add A LOT of annotations, and that is arguably starting to increase the complexity.

The CRD, with a K8s reconciler, would not only make it easier to add pluggable components, but would also (as discussed on Discord) have the much-needed benefit of allowing us to not rely on the Dapr Injector, which due to how K8s works, it's not reliable and can cause failures (hence why we had to create the "sidecar watchdog"). The DaprApp object would include a Deployment (or Job, etc), and the reconciler would then translate the DaprApp objects into Deployments/Jobs with daprd (and pluggable components) added.

(Of course, I'm not talking about completely deprecating support for using "regular" Deployment objects, but adding a second option which would be managed by a K8s operator/reconciler)

yaron2:
> This may not be the best place for this conversation, but reading this... Wouldn't it make sense to look into creating a specific CRD for Dapr-ized apps instead? Something like `DaprApp`?
> 
> We had a conversation about this on Discord a while back: https://discord.com/channels/778680217417809931/1033131188128317520/1033131188128317520
> 
> It's not that I'm against this proposal, which (with some changes perhaps as Yaron is discussing) is probably the best we can do right now. But we are starting to add A LOT of annotations, and that is arguably starting to increase the complexity.
> 
> The CRD, with a K8s reconciler, would not only make it easier to add pluggable components, but would also (as discussed on Discord) have the much-needed benefit of allowing us to not rely on the Dapr Injector, which due to how K8s works, it's not reliable and can cause failures (hence why we had to create the "sidecar watchdog"). The DaprApp object would include a Deployment (or Job, etc), and the reconciler would then translate the DaprApp objects into Deployments/Jobs with daprd (and pluggable components) added.
> 
> (Of course, I'm not talking about completely deprecating support for using "regular" Deployment objects, but adding a second option which would be managed by a K8s operator/reconciler)

As discussed, I support having a Dapr operator and I agree with you that it would provide a much better experience for Pluggable Components being first class citizens rather than annotation based, which is a concept that we are starting to abuse here.

I also agree with you that this discussion doesn't belong here, but rather in a separate proposal :)

ItalyPaleAle:
Only reason why I brought this up here is because I wonder if this makes sense to do anyways - or if we should look into the CRD option straight away

0005-R-pluggable-components-injector.md
yaron2: > it should be consistent with the runtime component list

What's the "runtime component list"? components are stored on on etcd in Kubernetes and are basically a shared global state. What race condition do you mean here?

> Use the daprClient to fetch the component list and apply the filter by scope logic

This should be the way to go, coupling the injector to the operator is extremely problematic for a number of reasons. I'm fine with this approach, the think the major disadvantage here is a `List` call to the k8s API server at scale. This is, frankly, the thing that worries me the most.

> I don't have a clear path here, it could be something like a components version that allow to fetch a specific version of the dapr configuration state, or something like we've discussed about having a init container that populates the components directory based on a base64-encoded components yaml.

Both paths here are a non-starter for me, but that's not relevant for now.

yaron2:
I left a few comments, but at this stage they are more implementation details and I feel like this proposal is in a good state.

+1 binding.

artursouza:
+1 I agree

yaron2:
There's binding votes majority, good to merge.

yaron2 (approved):

0005-R-pluggable-components-injector.md
mcandeia: > What's the "runtime component list"? components are stored on on etcd in Kubernetes and are basically a shared global state. What race condition do you mean here?

I meant this list => https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L2372

Try to imagine a scenario where the injector asks for all components in the current namespace in a timestamp T1, shortly after (before the target application pod gets running) the user adds a new pluggable component (i.e a new component yaml) T2, in a timestamp T3 the pod is ready to run and it will retrieve the list again from the operator using the [loader](https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L2372)

The list from runtime will be different from the list that the injector used when patching the pod.

As I said It could lead in two different situations:

1) inject a useless pod in case of T2 being a deletion of a component
2) do not inject a required pod that might cause panics

PR #12, opened 2022-11-23
Pubsub delayed scheduling proposal

yaron2:
Signed-off-by: yaron2 <schneider.yaron@live.com>

### DIFF
--- a/0005-R-pubsub-delayed-scheduling.md
+++ b/0005-R-pubsub-delayed-scheduling.md
@@ -1,0 +1,156 @@
+# Title of proposal 
+
+* Author: Yaron Schneider (@yaron2)
+* State: Ready for Implementation
+* Updated: 23/11/2022
+
+## Overview
+
+This is a design proposal for the widely requested [delayed message scheduling for pub/sub feature request](https://github.com/dapr/dapr/issues/2675).
+
+The goal of this feature enhancement is to provide developers with the ability to send messages to a pub/sub topic at a given, pre-configured time.
+
+## Background
+
+For background, see this [issue](https://github.com/dapr/dapr/issues/2675). Delayed message scheduling is a common feature in message brokers that helps in postponing essential workloads to a later time, where consumers can then pick them up either after peak hours or when the correct conditions for processing have been met. The use cases for delayed message sending are many, from delaying email sending to processing orders when enough computational capacity on the receiving end has been met to throttle prevention.
+
+### Related issues 
+
+https://github.com/dapr/dapr/issues/2675
+
+## Expectations and alternatives
+
+* What is in scope for this proposal?
+
+This proposal outlines a new metadata key for the [Dapr publish endpoint](https://docs.dapr.io/reference/api/pubsub_api/#publish-a-message-to-a-given-topic) to apply for both HTTP and gRPC APIs that allows developers to set the target date and time for a message to be published to a given topic. 
+
+* What is deliberately *not* in scope?
+
+Using component specific implementations is deliberately not in scope, as it offers no tangible benefits and carries heavy maintenance burden for contributors, approvers and maintainers.
+
+* What alternatives have been considered, and why do they not solve the problem?
+
+The only alternative for users is to create separate topics for delayed messages together with boilerplate code to orchestrate sending/receiving messages from said topics within their code, and then chaining the message flow to the correct topics after the conditions for sending delayed messages have been met. In addition, user code will need to become aware of Dapr specific mechanisms and mirror it in user code.
+
+* What advantages / disadvantages does this proposal have?
+
+This proposal allows delayed message scheduling to be enabled for:
+
+1. Every pub/sub component supported by Dapr without requiring component specific changes
+2. Messages sent using Dapr CloudEvents
+3. Messages sent using user provided CloudEvents
+4. Messages sent using raw payloads
+5. Message sent/received by HTTP and gRPC subscribers
+
+## Implementation Details
+
+### Design
+
+The Dapr runtime will use an intermediary topic to store messages that are scheduled for late delivery. Users will specify their intent for a message to be delayed using the existing metadata mechanism (specified as a query param on HTTP1.1 or on the request object for gRPC), the same way we use metadata for TTLs in pub/sub and for many other features. The proposed metadata key to be used is `scheduledSendTime` with the value being a date time string adhering to the ISO-8601 format, the same format supported by Dapr actors reminders and timers.
+
+HTTP example:
+
+```bash
+curl -X "POST" http://localhost:3500/v1.0/publish/pubsub/singularity?metadata.scheduledSendTime=2045-11-05T08:15:30-05:00 -H "Content-Type: application/json" -d '{"execute-order": "66"}'
+```
+
+Upon receiving a delayed message, the Dapr runtime will examine the due date for the message and publish it to the target topic if time is due. If the time isn't due, the Dapr runtime will hold the message until the time is right to send it without ACKing back to the broker. If the Dapr instance crashes, the message will remain in the broker and be consumed by a different instance.
+
+#### Configuring delayed messaging
+
+Users can enable delayed messaging by adding the following pub/sub component metadata (other fields specific for Redis removed for brevity):
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: pubsub
+spec:
+  type: pubsub.redis
+  version: v1
+  metadata:
+  - name: "enableDelayedMessaging"
+    value: true
+```
+
+If not set, the default value is `false` and the feature is bypassed completely.
+
+
+#### Consuming delayed messages
+
+On startup, the Dapr runtime will subscribe to a topic in the following format: `<app-id>-delayed`. This topic holds all delayed messages for the given application, and has the following benefits:
+
+1. Topics messages are scoped per application and not per component
+2. Enables infrastructure transparency in that operators can examine the underlying broker, discover the delayed messages topic and optionally look into the message queue if supported
+3. Of secondary importance, this enables users to optionally consume messages directly from the configured topic for advanced scenarios
+
+##### Interaction with namespaced consumer groups
+
+If [namespaced consumer groups](https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-namespace/#with-namespace-consumer-groups) are enabled, the topic format will result in `<app-id><namespace>-delayed` to allow for a delayed message topic that is scoped both to the app id and the namespace it is running in.
+
+#### Publishing delayed messages
+
+When Dapr receives a message that has a metadata key of `scheduledSendTime`, it will publish it to the delayed message topic in the format described above, if `enableDelayedMessaging` has been enabled on the component.
+Since this feature supports all pub/sub components, Dapr will return an error if a metadata key is present and `enableDelayedMessaging` is false or not set on the component.
+
+### Message format
+
+Dapr instances that receive a delayed message need to know the topic and other request level details (content type, etc.). To faciliate this requirement and support both CloudEvents and raw payloads, Dapr will publish a well known, typed proto message to the delayed messages topic in the following structure:
+
+```proto
+message DelayedMessage {
+  string topic = 1;
+  bytes message = 2;
+  string contentType = 3;
+  map<string, string> metadata = 4;
+  bool bulkPublish = 5;
+}
+```
+
+This provides the Dapr runtime all the information it needs to create a publish request and send it to the message broker.
+
+#### Bulk publish support
+
+If the `DelayedMessage` has the `bulkPublish` field set to `true`, the `message` field will be deserialized as a [BulkMessageEntry](https://github.com/dapr/components-contrib/blob/677663e81ab292d140021e671841ca5f613aaaea/pubsub/requests.go#L26) type and will be sent via the bulk publish route.
+
+#### Components with topic management disabled
+
+Some Dapr pub/sub components allow users to override the creation of topics. In that case, documentation will state to users they need to pre-create the delayed message topic using the correct format.
+
+### Feature lifecycle outline
+
+* Expectations
+
+This feature will start as a preview feature.
+
+* Compatibility guarantees
+
+This feature is backward compatible with previous Dapr versions.
+
+* Feature flags
+
+This feature requires the Preview feature flag.
+
+### Acceptance Criteria
+
+How will success be measured? 
+
+* Performance targets
+
+N/A
+
+* Compabitility requirements
+
+Implementation needs to be compliant with previous versions of Dapr when bypassed (disabled) in terms of API endpoints, behavior, payloads and error codes.
+
+* Metrics
+
+N/A
+
+## Completion Checklist
+
+What changes or actions are required to make this proposal complete?
+
+* Code changes
+* Tests added (e2e, unit)
+* Documentation
--- a/0005-R-pubsub-delayed-scheduling.md
+++ b/0005-R-pubsub-delayed-scheduling.md
@@ -1,4 +1,4 @@
-# Title of proposal 
+# Pubsub delayed scheduling
 
 * Author: Yaron Schneider (@yaron2)
 * State: Ready for Implementation


0005-R-pubsub-delayed-scheduling.md
```diff
@@ -0,0 +1,156 @@
...
+  bytes message = 2;
+  string contentType = 3;
+  map<string, string> metadata = 4;
+  bool bulkPublish = 5;
+}
+```
+
+This provides the Dapr runtime all the information it needs to create a publish request and send it to the message broker.
+
+#### Bulk publish support
```
yaron2, line 112: @mukundansundar please verify this is feasible.

mukundansundar: A `BulkPublishRequest` contains multiple `BulkMessageEntry` entries. 
So if my understanding of this section is correct, you want to store the byte serialized form `BulkMessageEntry` in the` message` field in `DelayedMessage`. If that is the case, why store the `BulkMessageEntry` object? Can't we not deserialize when publishing to the delayed intermediate topic and only store the data and metadata associated with it? 

The main part of the BulkPublish API is the optimization between App and Dapr and Component Service. 

Even if the component does not support native BulkPublish support, can we send all delayed messages as a BulkPulblish request making use of the default BulkPublish route?

yaron2: > Even if the component does not support native BulkPublish support, can we send all delayed messages as a BulkPulblish request making use of the default BulkPublish route?

So that's the idea, once Dapr has received a delayed message (that contains all message entries), it will publish it using the BulkPublish route.

0005-R-pubsub-delayed-scheduling.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+If not set, the default value is `false` and the feature is bypassed completely.
+
+
+#### Consuming delayed messages
+
+On startup, the Dapr runtime will subscribe to a topic in the following format: `<app-id>-delayed`. This topic holds all delayed messages for the given application, and has the following benefits:
+
+1. Topics messages are scoped per application and not per component
```
mcandeia, line 83: needs clarification:

I'm assuming that some component must be configured at first to be the owner of the `-delayed` topics, correct me if I'm wrong,

so my questions are:

1) How this will be configured?
2) Can the broker send those messages to a dead-letter topic due to the lack of ACKing? If yes, will Dapr subscribe to a dead letter topic while not ACKing the messages?  If not, should we have a way to distinguish between a runtime bug that prevents ACKing messages, and the desired behavior?
3) given that the topic is not component-scoped, what would happen in case of the target pubsub component has been deleted while trying to publish the message ?

0005-R-pubsub-delayed-scheduling.md
```diff
@@ -0,0 +1,156 @@
...
+
+The Dapr runtime will use an intermediary topic to store messages that are scheduled for late delivery. Users will specify their intent for a message to be delayed using the existing metadata mechanism (specified as a query param on HTTP1.1 or on the request object for gRPC), the same way we use metadata for TTLs in pub/sub and for many other features. The proposed metadata key to be used is `scheduledSendTime` with the value being a date time string adhering to the ISO-8601 format, the same format supported by Dapr actors reminders and timers.
+
+HTTP example:
+
+```bash
+curl -X "POST" http://localhost:3500/v1.0/publish/pubsub/singularity?metadata.scheduledSendTime=2045-11-05T08:15:30-05:00 -H "Content-Type: application/json" -d '{"execute-order": "66"}'
+```
+
+Upon receiving a delayed message, the Dapr runtime will examine the due date for the message and publish it to the target topic if time is due. If the time isn't due, the Dapr runtime will hold the message until the time is right to send it without ACKing back to the broker. If the Dapr instance crashes, the message will remain in the broker and be consumed by a different instance.
```
mcandeia, line 57: should we worry about it becoming a bottleneck? like having a large number of messages held and trying to be delivered but in a "n-ack" loop for any reason?

0005-R-pubsub-delayed-scheduling.md
yaron2: > How will this be configured?

This is explained below.. see [here](https://github.com/dapr/proposals/pull/12/files#diff-facc607b55c1491e99f88ba1b46ba1c3b3fa8841614b300a6a938239bf3c6d09R59).

> Can the broker send those messages to a dead-letter topic due to the lack of ACKing?

That would be undesirable. Dead letters topics are set per topic and handle dead lettering for the application processing.

> If not, should we have a way to distinguish between a runtime bug that prevents ACKing messages, and the desired behavior?

Delayed messages shouldn't prevent ACKing messages for the end topic, once a delayed message has been published to the end topic it will be ACKed by the app. The runtime will only ACK the message for the delayed message topic after it has been sent to the target topic successfully. Failures to ACK back to the delayed message topic will show up in the logs with the name of the delayed message topic in the same flow we have today.

> given that the topic is not component-scoped, what would happen in case of the target pubsub component has been deleted while trying to publish the message ?

The same thing that happens if a pubsub component is deleted while trying to publish a message to any other topic, be it an app topic, deadletter or otherwise - Dapr returns an error to the user code.

0005-R-pubsub-delayed-scheduling.md
yaron2: Messages will be distributed between N instances of Dapr running, and I don't see this becoming a bottleneck any more than bulk pub/sub. Given that Dapr publishes these messages to the pub/sub and not the app likely guarantees more consistent throughput. This is more of an implementation detail, but there's a lot of flexibility we can introduce further down the road if needed to only hold a certain number of messages in an instance and NACK the others back to the message bus.

0005-R-pubsub-delayed-scheduling.md
mcandeia: > This is explained below.. see [here](https://github.com/dapr/proposals/pull/12/files#diff-facc607b55c1491e99f88ba1b46ba1c3b3fa8841614b300a6a938239bf3c6d09R59).

I was thinking they were different things. One thing is which pubsub will be used to store delayed messages and another thing is whether the pubsub in question supports delayed messages or not. So the `enableDelayedMessaging` has the same semantics as `actorStateStore` for statestore buildingblocks? is that correct?

> That would be undesirable. Dead letters topics are set per user topic and handle dead lettering for the application processing.
Got it,
Just a heads up to make it clear through documentation, I was thinking about specific configurations that might be needed in case of manually creating such `-delayed` topics

> Delayed messages shouldn't prevent ACKing messages for the end topic, once a delayed message has been published to the end topic it will be ACKed by the app.

I'm trying to imagine a scenario where the target pubsub is out for any reason, probably this would lead in n-ACKing messages by dapr runtime, not because its part of the message delay, but because there are something preventing the message to be delivered to the target topic, and, because of this, the message retention would increase indefinitely.

yaron2: > another thing is whether the pubsub in question supports delayed messages or not

Please read the proposal fully. All pub subs support delayed messages.

yaron2: > I was thinking about specific configurations that might be needed in case of manually creating such -delayed topics

There is such a case, explained here: https://github.com/dapr/proposals/pull/12/files#diff-facc607b55c1491e99f88ba1b46ba1c3b3fa8841614b300a6a938239bf3c6d09R118

yaron2: > I'm trying to imagine a scenario where the target pubsub is out for any reason, probably this would lead in n-ACKing messages by dapr runtime, not because its part of the message delay, but because there are something preventing the message to be delivered to the target topic, and, because of this, the message retention would increase indefinitely.

As you mentioned, this is unrelated to delayed scheduling.

mcandeia: > Please read the proposal fully. All pub subs support delayed messages

I understand that this functionality is provided by the runtime. My question was if the flag would serve as a feature-flag to enable/disable for a specific pubsub or if it would serve to mark a pubsub as the owner of the `-delayed` topics but you answred it, thanks.

> There is such a case, explained here

I saw that. Just emphasizing that we should address DLQ's specifically

> As you mentioned, this is unrelated to delayed scheduling.

Yup. But it doesn't mean that we can't do anything about it, for example we can have a TTL or a maximum number of retries for attempted delivery, does that make sense?

yaron2: > My question was if the flag would serve as a feature-flag to enable/disable for a specific pubsub or if it would serve to mark a pubsub as the owner of the -delayed topics but you answred it, thanks.

This feature will have a preview feature flag regardless, but there's the specified specific field that explicitly enables delayed messages for a component.

yaron2: > Yup. But it doesn't mean that we can't do anything about it, for example we can have a TTL or a maximum number of retries for attempted delivery, does that make sense?

This kind of overlaps with pub/sub resiliency, but this can and should be discussed in dapr/dapr.

mcandeia: Just to correct myself, because I said "delivered" but actually the "delivered" for this message is a "publish" in the target pubsub. 

I understand that the resiliency is more "stateless" than we need for that, but I could be talking nonsense. When I mentioned "retry" I meant number of retries even between different runtimes (in case of a crash for instance) and I don't know how our resiliency policy would currently handle TTL (sorry). But I think the subject in general is about resiliency yes, I just don't know if it would fit into something related to pubsub to be honest.

0005-R-pubsub-delayed-scheduling.md
```diff
@@ -0,0 +1,156 @@
...
+## Implementation Details
+
+### Design
+
+The Dapr runtime will use an intermediary topic to store messages that are scheduled for late delivery. Users will specify their intent for a message to be delayed using the existing metadata mechanism (specified as a query param on HTTP1.1 or on the request object for gRPC), the same way we use metadata for TTLs in pub/sub and for many other features. The proposed metadata key to be used is `scheduledSendTime` with the value being a date time string adhering to the ISO-8601 format, the same format supported by Dapr actors reminders and timers.
+
+HTTP example:
+
+```bash
+curl -X "POST" http://localhost:3500/v1.0/publish/pubsub/singularity?metadata.scheduledSendTime=2045-11-05T08:15:30-05:00 -H "Content-Type: application/json" -d '{"execute-order": "66"}'
```
daixiang0, line 54: My concern is that what is the standard of time, client, dapr or other? Is there any potential time diff issue?

yaron2: The supported datetime format here is ISO-8601 (same one we use today for creating Dapr reminders and timers), which in the example above is:

```2045-11-05T08:15:30-05:00```

-05:00 means it's in eastern standard timezone UTC. If a timezone isn't specified, Dapr is the standard of time, which in almost all cases would be the same as the client as they are running on the same machine.

Comment (daixiang0) [reply 0005-R-pubsub-delayed-scheduling.md:54]: It is meaningful.

**Review by daixiang0 — DISMISSED** [review]
LGTM

**Comment by daixiang0** [comment]
+1 binding

**Comment by yaron2** [comment]
+1 binding

File: 0005-R-pubsub-delayed-scheduling.md [hunk 5]
```diff
@@ -0,0 +1,156 @@
...
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: pubsub
+spec:
+  type: pubsub.redis
+  version: v1
+  metadata:
+  - name: "enableDelayedMessaging"
+    value: true
```
mukundansundar, line 73: When this flag is set to true, we will be using the same `pubsub` with a different topic i.e. `<app>-delayed` to store the delayed messages and then later process it and send it to the correct topic in the `pubsub`?

0005-R-pubsub-delayed-scheduling.md
```diff
@@ -0,0 +1,156 @@
...
+  map<string, string> metadata = 4;
+  bool bulkPublish = 5;
+}
+```
+
+This provides the Dapr runtime all the information it needs to create a publish request and send it to the message broker.
+
+#### Bulk publish support
+
+If the `DelayedMessage` has the `bulkPublish` field set to `true`, the `message` field will be deserialized as a [BulkMessageEntry](https://github.com/dapr/components-contrib/blob/677663e81ab292d140021e671841ca5f613aaaea/pubsub/requests.go#L26) type and will be sent via the bulk publish route.
```
mukundansundar, line 114: Why is this being treated differently? 
Instead can this following flow be done? 
When a bulkPublish request is made and scheduledSendTime is set on any message in that request, that message with metadata alone is sent to the `<app>-delayed` topic, why do we need to send the whole of the BulkMessageEntry to the topic? 
On subscribing to the `<app>-delayed` topic, only the message with metadata will be there in terms of cloudevent or for raw payload only the message payload, that alone can then be forwarded to the correct topic.

Main  question is in the `<app>-delayed` topic, why should we serialize and store BulkMessageEntry? Why can't we just store the Dapr created CE/ User input CE/ raw payload directly even for bulkPublish?

mukundansundar (commented):
One more question @yaron2 is what happens to the `publishingScopes` and `subscribingScopes` definition that we have for limiting topic access? https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-scopes/ 

How will that work with `<app>-delayed` topic?

0005-R-pubsub-delayed-scheduling.md
yaron2: Yes, only if a publish reauest containing the metadata key for a scheduled event is present. Otherwise it's delivered to the target topic as is.

0005-R-pubsub-delayed-scheduling.md
yaron2: We could, but looking at the code currently in master branch it looks like the publish bulk method in contrib expects a BulkMessageEntry. 
This is more of an implementation detail.

yaron2:
> One more question @yaron2 is what happens to the `publishingScopes` and `subscribingScopes` definition that we have for limiting topic access? https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-scopes/
> 
> How will that work with `<app>-delayed` topic?

It works as is, unchanged. Again, this is an implementation detail but Dapr checks if the app is allowed to publish before it sends it to any topic, delayed or not. On the receiving side, Dapr checks if the app is allowed to subscribe regardless of delayed topics as they only act as an intermediary.

mukundansundar:
> > One more question @yaron2 is what happens to the `publishingScopes` and `subscribingScopes` definition that we have for limiting topic access? https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-scopes/
> > How will that work with `<app>-delayed` topic?
> 
> It works as is, unchanged. Again, this is an implementation detail but Dapr checks if the app is allowed to publish before it sends it to any topic, delayed or not. On the receiving side, Dapr checks if the app is allowed to subscribe regardless of delayed topics as they only act as an intermediary.

The scoped topics won't change but in addition to that will the application need access to an `<app>-delayed` topic?

yaron2:
> > > One more question @yaron2 is what happens to the `publishingScopes` and `subscribingScopes` definition that we have for limiting topic access? https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-scopes/
> > > How will that work with `<app>-delayed` topic?
> > 
> > 
> > It works as is, unchanged. Again, this is an implementation detail but Dapr checks if the app is allowed to publish before it sends it to any topic, delayed or not. On the receiving side, Dapr checks if the app is allowed to subscribe regardless of delayed topics as they only act as an intermediary.
> 
> The scoped topics won't change but in addition to that will the application need access to an `<app>-delayed` topic?

No, this is a Dapr internally used topic.

0005-R-pubsub-delayed-scheduling.md
artursouza, line 57: This will not work in every broker. Not acking a message can cause no other messages to be consumed in some configurations. In Kafka, there is in-order guarantee per partition, so not acking a message will clog the partition for other messages - even for those are already due.

Instead, this should implement an outbox pattern instead. In the outbox pattern, messages are only published after a business transaction takes place - in this case, the business transaction is done by the clock. Then, the component must scan states in a state store to identify which messages are due and publish those, deleting the corresponding record.

The state store keys would basically be in segments of 1h (or any other granularity) and contain the ids for every message that must be delivered in that hour. Then, daprd would look at the current hour and previous (configurable by the user to be -n hours) and publish every message that is due in those time windows.

This will make this solution work in a predictable way without depending on broker-specific handling of "not acking messages" by just relying on the state store pattern underneath.

artursouza (changes requested):

ItalyPaleAle:
I second @artursouza's concerns that it may cause some issues with a variety of brokers. Artur mentioned Kafka. Two more examples:

- Azure Service Bus requires consumers to acquire and constantly renew a lock, or the message will automatically be nack'd. Renewing locks can become expensive when we have a lot of messages that are delayed.
- MQTT's relationship with ACK's is..... [complex](https://github.com/dapr/components-contrib/issues/2329), to say the least.
- Some brokers automatically delete messages that have been in the queue for longer than N days and have not been ack'd.

I also have a question about what happens if the consumer is scaled horizontally. Among all PubSub components we support right now, we observe a variety of behaviors, but many (not all) deliver all messages to all subscribers. When the app is scaled horizontally, then, how do we coordinate this?

I definitely do see the value in having delayed messages and it's a big pain point. However, I share the concerns with this proposed implementation's feasibility (and long term supportability across all brokers). Ideally, we should encourage users to rely on the native support for delayed messages that brokers (like Service Bus) offer whenever possible.

Alternatively, if we need a Dapr-specific implementation, I would consider something that uses a state store to temporarily persist messages until they're ready to be published.

artursouza:
I would also consider making the state store bound to a particular topic - appID combination to avoid contention between multiple apps.

yaron2:
> When the app is scaled horizontally, then, how do we coordinate this?

Coordination isn't needed as this would satisfy the at-least-once guarantee.

The concern about queue clogging is real and using a general purpose state store makes sense, I'll update the proposal for that section.

0005-R-pubsub-delayed-scheduling.md
berndverst: @yaron2 has this comment been integrated yet? If so can you resolve the conversation :)

yaron2: It hasn't been resolved yet

olitomlinson:
@yaron2 

Personally, I would advocate utilise Dapr Workflows to handle the time-bound aspects of business logic, and not push that responsibility into the PubSub broker.

image

---

_Assuming that a [DurableTimers](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-timers?tabs=csharp) equivalent make it into the Dapr embedded workflow API surface_ 

**Example**

PubSub Subscriber starts a workflow instance for each message.

```c#
// Dapr subscription in [Topic] routes orders topic to this route
app.MapPost("/orders", [Topic("orderpubsub", "orders")] (Order order) => {
    await workflowClient.ScheduleNewWorkflowAsync(
        name: nameof(OrderProcessingWorkflow),
        instanceId: order.orderId,
        input: order);
    return Results.Ok(order);
});
```

Use a workflow to host the logic, put use a Timer to defer the work until the desired time, which could be relative to a property in the message, or a fixed window etc.

```c#
class OrderProcessingWorkflow : Workflow<OrderPayload, OrderResult>
{
    public override async Task<OrderResult> RunAsync(WorkflowContext context, OrderPayload order)
    {
        // Put the workflow to sleep for 72 hours because thats how the business wants to handle this message type
        DateTime dueTime = context.CurrentUtcDateTime.AddHours(72);
        await context.CreateTimer(dueTime, CancellationToken.None);

        // Notify the user that an order has come through
        await context.CallActivityAsync(
            nameof(NotifyActivity),
            new Notification($"Received order {orderId} for {order.Quantity} {order.Name} at ${order.TotalCost}"));

       ....
    }
}
   ```

kayvonkhosrowpour:
+1 for pubsub publishing with delay

ItalyPaleAle:
What if we implemented this on top of actors instead?

We could leverage the actor reminder subsystem to publish messages at a specific time. We'd also get "for free" the ability to publish messages with a delay.

yaron2:
> What if we implemented this on top of actors instead?
> 
> We could leverage the actor reminder subsystem to publish messages at a specific time. We'd also get "for free" the ability to publish messages with a delay.

That's an idea I've been looking at recently, but would have to defer this to reminders v2. We're planning a distributed scheduling solution in Dapr to serve higher level APIs like actor reminders, delayed pub/sub and an upcoming outbox pattern API

ItalyPaleAle:
Yes, agree "reminders v1" would likely not offer the perf this would require.

However, it would be nice to see how this could look like implemented on top of an internal actor, like workflows, even if the implementation weren't be possible until "reminders v2".

olitomlinson:
@yaron2 Does a public forum exist for discussing the 'new distributed scheduling solution' ?

ulf-melin-sveasolar:
What is happening to this?

yaron2:
> What is happening to this?

We are working on a distributed scheduling system for Dapr which among other consumers (Actors and Workflows) could also be used to support this delayed messaging feature. This proposal will be updated once we have the new system in place

PR #14, opened 2022-11-29
Building Block: Proposal document for Gatekeeper Building Block

XavierGeerinck:
Signed-off-by: Xavier Geerinck <xavier.geerinck@gmail.com>

### DIFF
--- a/0007-B-gatekeeper-building-block.md
+++ b/0007-B-gatekeeper-building-block.md
@@ -1,0 +1,272 @@
+# Title of proposal 
+
+* Author(s): Xavier Geerinck
+* State: Ready for Implementation
+* Updated: 29/NOV/2022
+
+## Overview
+
+This proposal documents introduces the "Gatekeeper" building block which helps developers to implement "Authorization" requests for their application, assisting the developer by implementing core authorization requirements:
+* Am I as a user able to access resource X?
+* Which resources am I as a user able to access? (commonly known as an "ARQ" or "Access Request Query")
+
+It works through added APIs in the Dapr runtime that work together with a state store to store the policies.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


XavierGeerinck:
@dapr/maintainers-dapr thoughts?

yaron2:
@XavierGeerinck will be reviewing this soon, as back for vacation now.

jogobella00:
That would be great to have it in Dapr already, would it be possible to have this PR approved?

Yunzlez:
+1 Also waiting for this!

artursouza:
Can it follow a spec 100%, maybe Cerbos? Similar to how we have 100% compatible cloud event in pubsub. Then, show how this can be integrated with existing 3rd party  frameworks. I think the biggest value add if it can interop outside the Dapr SDKs.

XavierGeerinck:
I think it would indeed allow for a 100% compatibility with their spec, the question is just what the added value of Dapr would be then. Integrating it natively in Dapr could be more seen as building block, while a cloud event is more seen as interop layer for integration

berndverst:
Can someone copy the proposal text also into the description here so we can discuss this better?

Also, I'm hesitant for us to merge proposals without any identified stakeholders who are committed to driving the engineering work / implementation.

XavierGeerinck:
> Can someone copy the proposal text also into the description here so we can discuss this better?
> 
> Also, I'm hesitant for us to merge proposals without any identified stakeholders who are committed to driving the engineering work / implementation.
> 
> I don't fully see / understand the value of this proposal for Dapr. An interop layer makes more sense to me.

Done (partly though as it makes no sense for an entire duplication)

The ideation (and community request - see references in the proposal) are around providing an microservice angostic way of authorizing calls. The value for Dapr is immense as it is a component required by many people and would even improve Daprs current authorization mechanisms.

An interop layer could work as well, but this is where the discussion comes in in if Dapr wants this as USP (extra building block type) or not.

PR #15, opened 2022-11-30
Create PR 0011-R-secret-auto-injection.md

ezYakaEagle442:
Signed-off-by: Steve aka ezYakaEagle442 <43554247+ezYakaEagle442@users.noreply.github.com>

### DIFF
--- a/0011-R-secret-auto-injection.md
+++ b/0011-R-secret-auto-injection.md
@@ -1,0 +1,198 @@
+# Title of proposal 
+
+* Author(s): Steve P.
+* State: Ready for Implementation
+* Updated: 23/11/2022
+
+## Overview
+
+A brief description of the proposal; include information such as:
+
+* What areas are affected by this change? ==> Runtime & eventually SDKs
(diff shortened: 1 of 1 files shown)


ItalyPaleAle:
Thanks for this proposal and for taking the time to write it out!

I am not sure I support this proposal however. The main issue I have is that for this to work, we'd need the Dapr Sidecar Injector to communicate with secret stores, which is currently something we do not do and we'd have a hard time supporting: issues include authorization, scoping, and performance. I also think that there will be some UX issues, for example with the fact that secrets that are injected by the injector are not updated during the app's lifecycle, and the next natural question we'd get is how to allow hot-reload.

yaron2:
-1 binding.

This proposal introduces both security unwarranted security implications (sidecar reading application, non-Dapr secrets), broadening the scope of the injector to handle non-Dapr issues and requiring the injector to know intimate details regarding the application architecture.

artursouza:
-1 binding

Language specific details should fall into the SDK. Although it is nice to have an user experience where Dapr can be used without requiring *any* code changes, I don't think it is possible in this case (see Yaron's comments). The next best thing is for the app to bring in a Dapr SDK dependency that works really well with their framework of choice and require zero or a few code/config changes.

PR #16, opened 2022-12-09
Proposal: serverless compute target

ItalyPaleAle:
This is a proposal for being able to start a Dapr runtime ("daprd") as a proxy to **serverless compute** resources, a collective term that includes things like Azure Functions, AWS Lambda, Cloudflare Workers (and workerd), OpenFaaS, etc.

The goal is to allow developers to build apps that can leverage serverless compute in a smooth way, transparently integrating with Dapr. It allows building apps that can leverage flexible, on-demand compute resources, as a target for Dapr service invocation, PubSub messages, or input bindings.

This will be implemented by allowing Dapr to use a new class of components, called "compute", as a target for Dapr API calls, instead of an app running side-by-side with the Dapr runtime.

### DIFF
--- a/0009-BCR-serverless-target.md
+++ b/0009-BCR-serverless-target.md
@@ -1,0 +1,161 @@
+# Serverless compute target
+
+* Author(s): @ItalyPaleAle
+* State: Draft
+* Updated: 2022-12-09
+
+## Overview
+
+This is a proposal for being able to start a Dapr runtime ("daprd") as a proxy to **serverless compute** resources, a collective term that includes things like Azure Functions, AWS Lambda, Cloudflare Workers (and workerd), OpenFaaS, etc.
+
+The goal is to allow developers to build apps that can leverage serverless compute in a smooth way, transparently integrating with Dapr. It allows building apps that can leverage flexible, on-demand compute resources, as a target for Dapr service invocation, PubSub messages, or input bindings.
+
+This will be implemented by allowing Dapr to use a new class of components, called "compute", as a target for Dapr API calls, instead of an app running side-by-side with the Dapr runtime.
+
+## Background
+
+Serverless compute offers a programming model that allows developers to run code (usually written in JavaScript/TypeScript or compiled to WebAssembly) inside a "serverless runtime", which includes cloud-hosted services (Azure Functions, AWS Lambda, Cloudflare Workers, etc) as well as self-hosted projects (workerd, OpenFaaS). The main benefit to developers is that it expose compute resources that are flexible, available on-demand, and "infinitely scalable", so they can also tolerate workloads that are highly burstable.
+
+With this proposal, we will make it easier for developers to add code running on serverless compute and integrate that seamlessly in an architecture that is based on Dapr. Developers will be able to make service invocation calls from other Dapr-ized apps that are served by serverless compute, as well as use serverless compute resources for processing PubSub messages and input bindings via Dapr.
+
+## Expectations and alternatives
+
+* **What is in scope for this proposal?**
+  - This is a proposal for starting a Dapr sidecar that targets a serverless app for all messages sent from Dapr to the app. It will allow creating pods where only daprd is running and targets a serverless function for service invocation and for delivering PubSub messages and input bindings.
+  - For each serverless platform we support, we will need to create either a "SDK" or a "gateway function" (depending on the capabilities of the platform) that will take care of the "plumbing" to allow Dapr to invoke the serverless code the developers write, taking care of things like authentication (see the "implementation details" section for more details).
+* **What is deliberately *not* in scope?**
+  - This is not a proposal for integrating Dapr in a serverless environment.
+  - Although this will allow Dapr to invoke an app that is running as serverless, it does not allow the app to call into the Dapr APIs. This is because it's expected that Dapr will reside in places such as a K8s cluster and will be behind a firewall, while serverless code often runs in managed cloud environments, so communication from the app to Dapr won't be possible. Additionally, only HTTP can be used for the app channel.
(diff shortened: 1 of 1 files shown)


olitomlinson:
> This is a proposal for being able to start a Dapr runtime ("daprd") as a proxy to serverless compute resources, a collective term that includes things like Azure Functions, AWS Lambda, Cloudflare Workers (and workerd), OpenFaaS, etc.

I've read the proposal and I'm unsure about why there is so much emphasis on the word 'serverless' ? 

I can only assume that because of the emphasis on serverless, It won't be possible to proxy onto any traditional web server that exposes HTTP endpoints?

To me the underlying host technology (serverless or otherwise) is not important if daprd is _just_ invoking HTTP endpoints on the target?

--- 

why _must_ the CRD need to know about the external computes infrastructure via the `Type ` attribute i.e `compute.azure.functions` ?  For example, why would daprd operate differently if the target was an Azure Function vs an AWS Lambda?

yaron2:
> > This is a proposal for being able to start a Dapr runtime ("daprd") as a proxy to serverless compute resources, a collective term that includes things like Azure Functions, AWS Lambda, Cloudflare Workers (and workerd), OpenFaaS, etc.
> 
> I've read the proposal and I'm unsure about why there is so much emphasis on the word 'serverless' ?
> 
> I can only assume that because of the emphasis on serverless, It won't be possible to proxy onto any traditional web server that exposes HTTP endpoints?
> 
> To me the underlying host technology (serverless or otherwise) is not important if daprd is _just_ invoking HTTP endpoints on the target?
> 
> why _must_ the CRD need to know about the external computes infrastructure via the `Type ` attribute i.e `compute.azure.functions` ? For example, why would daprd operate differently if the target was an Azure Function vs an AWS Lambda?

Your feedback is spot on. There's nothing "serverless" about the features and specification pointed out here and there's no common ground between them. This is basically about adding an authentication mechanism to external, non-Dapr HTTP endpoints, and is really a nice extension to my proposal [here](https://github.com/dapr/dapr/issues/4549) which augments it.

There may be differences in how to authenticate or call an Azure Function or AWS Lambda, but the differences are semantics of an HTTP endpoint, not a serverless platform.

To illustrate this further:

1. An organization wants to use service invocation to an external proprietary system via HTTP and has specific authentication
2. An organization wants to use service invocation to invoke any HTTP cloud service that's not considered "serverless", ie. Azure App Service, AWS Elastic Beanstalk etc
3. An organization wants to use service invocation to invoke an API gateway with custom auth, whether internal or external to their org
4. An organization wants to use service invocation to invoke a non-Dapr HTTP service on Kubernetes/VM

These examples are very common use cases and non are considered serverless, yet all of them could be enabled if Dapr could invoke external non-Dapr endpoints and had the proper authentication configuration to do so, which this proposal covers well.

So my main feedback here is to join this proposal with mine (which I am going to transfer here from https://github.com/dapr/dapr/issues/4549) as it adds important missing pieces to it. I'm open to co-authoring @ItalyPaleAle.

ItalyPaleAle:
Thanks for the feedback.

I think dapr/dapr#4549 has some similarities but they are trying to solve different problems. I don't think they can be bundled together, to be honest.

The core part of this proposal is the change to the Configuration CRD to have Dapr initialize an app channel to an arbitrary HTTP(S) endpoint. It will still require a `daprd` process for each app, but it won't require a locally-running (ie. in the same pod) app and `daprd` will initialize an app channel towards an external HTTP(S) endpoint.

Here's how I see the two proposals, side-by-side:

| dapr/dapr#4549                                                 | This proposal                                                                                                          |
|----------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------|
| Only involves service invocation                               | Applies to service invocation and inbound pubsub and binding messages                                                  |
| Dapr on caller app invokes the external endpoint directly      | (For service invocation) A daprd runtime is required on the callee side too. The caller daprd invokes the callee daprd |
| High degree of flexibility for making service invocation calls | Very low degree of flexibility: target serverless code expects inputs received in a very specific format               |
| Goal is to allow invoking endpoints outside of Dapr            | Goal is to allow developers to scale their apps by leveraging "serverless" compute resources                           |

In this case, it's sort of required for the target to be a serverless compute endpoint, if not technically at least "philosophically" so it achieves the goals explained above.

yaron2:
> Thanks for the feedback.
> 
> I think [dapr/dapr#4549](https://github.com/dapr/dapr/issues/4549) has some similarities but they are trying to solve different problems. I don't think they can be bundled together, to be honest.
> 
> The core part of this proposal is the change to the Configuration CRD to have Dapr initialize an app channel to an arbitrary HTTP(S) endpoint. It will still require a `daprd` process for each app, but it won't require a locally-running (ie. in the same pod) app and `daprd` will initialize an app channel towards an external HTTP(S) endpoint.
> 
> Here's how I see the two proposals, side-by-side:
> 
> [dapr/dapr#4549](https://github.com/dapr/dapr/issues/4549)	This proposal
> Only involves service invocation	Applies to service invocation and inbound pubsub and binding messages
> Dapr on caller app invokes the external endpoint directly	(For service invocation) A daprd runtime is required on the callee side too. The caller daprd invokes the callee daprd
> High degree of flexibility for making service invocation calls	Very low degree of flexibility: target serverless code expects inputs received in a very specific format
> Goal is to allow invoking endpoints outside of Dapr	Goal is to allow developers to scale their apps by leveraging "serverless" compute resources
> In this case, it's sort of required for the target to be a serverless compute endpoint, if not technically at least "philosophically" so it achieves the goals explained above.

I still think it makes sense to combine the proposals into one, but regardless of https://github.com/dapr/dapr/issues/4549 at all, I would strongly advise to focus this proposal on the ability to have Dapr invoke non-localhost HTTP endpoints for service invocation, bindings and pub/sub and remove any mention of serverless all-together as there are no serverless semantics here, just HTTP + custom auth ones. Until then, this is a showstopper for me.

olitomlinson:
I agree, So far...I don't see how the target being serverless or not is relevant to the proposal?

PR #19, opened 2023-02-18
In-tree pluggable components

ItalyPaleAle:
Formalizing a proposal [initially discussed on Discord](https://discord.com/channels/778680217417809931/1067116086337745076)

This is a proposal for implementing the infrastructure to allow in-tree components to be shipped with Dapr, in the same Docker image as daprd, but instantiated as pluggable components by the Dapr runtime.

This is **not** a proposal to remove components from Dapr's distribution or remove them from the Dapr repositories. Components that are instantiated as in-tree pluggable components should work for end-users in the same way that statically-compiled components do.

### DIFF
--- a/000X-R-In-tree-pluggable-components.md
+++ b/000X-R-In-tree-pluggable-components.md
@@ -2,7 +2,7 @@
 
 * Author(s): Alessandro Segala (@ItalyPaleAle)
 * State: {Ready for Implementation, Implemented}
-* Updated: [Date]
+* Updated: 2023-02-18
 
 ## Definitions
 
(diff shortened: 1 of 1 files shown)


000X-R-In-tree-pluggable-components.md
```diff
@@ -0,0 +1,132 @@
...
+* Effect of security vulnerabilities in third-party packages.  
+  Most of our components depend on third-party packages such as SDKs, and we import dozens of them in Dapr. Suppose that component X depended on a library for which a vulnerability was found. Users who do not leverage component X are likely not impacted by the vulnerability, but they are still running a binary that depends on a vulnerable package, possibly causing compliance conerns.
+* Ability to ship some components with outsized impact on performance.  
+  Already in the last months we've had components that were accepted in `dapr/components-contrib`, after being reviewed, and then we had to remove them because we noticed they had an unacceptable impact on the resources used by `daprd`. Sadly, our processes and tools don't allow us to catch these things easily until the component is actually registered in `daprd`.
```
tmacam, line 33: It would be belter to anchor this time reference by specifically mentioning the last months of 1.10 release development. "In the last months" will be ambiguous 6 months from now.

000X-R-In-tree-pluggable-components.md
```diff
@@ -0,0 +1,132 @@
...
+  * Removing components from Dapr entirely is an option as well, for example to ship them as external, pluggable components. These would be separate from `daprd` in all ways, including: they're shipped in separate Docker images, need to be explicitly added to applications (see [0005-R-pluggable-components-injector.md](./0005-R-pluggable-components-injector.md)), and can have separate release lifecycles and support.
+  * Users can also re-compile Dapr with a custom set of components. Although we don't officially support this, we are aware of at least some users who do this as a practice. This is something only advanced users may consider doing due to the very high complexity and operational burden.
+* **Are there any trade-offs being made?**  
+  * Components that are instantiated as in-tree pluggable components will likely have *slightly* degraded performance, due to the overhead of inter-process communication and gRPC serialization/unserialization. Based on the research that was done for pluggable components, that should be negligible for most users.
```
tmacam, line 58: It woud be great if we could quantify this but I don't think we ever came with those numbers in  pluggable components final incarnation.

000X-R-In-tree-pluggable-components.md
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+> The code above is to be considered "pseudo-code" for reference only. The acutal interfaces of the internal packages are implementation details outside of the scope of this proposal.
+
+The component is then compiled like any other Go binary, creating a statically-compiled binary `intree_pluggable`.
+
+### Packaging
+
+The last step is packaging. After building all binaries, **they are included in the same container image as `daprd`**, for example in the `/components/` path within the container.
```
daixiang0, line 118: If in this way, the image size is still big.

I have an idea that when dapr init, download all biniaries at each machine, `daprd` mount that dir when run. End users still deploy as before.

ItalyPaleAle: I don't think we have concerns with the size of the image. What is most concerning to us is the size of the binary when is loaded in memory.

I am not a fan of downloading binaries on each machine, because:

1. It will make each container startup slower since the binaries need to be fetched over the network. If they're included in the Docker image, they are generally cached in the host already.
2. On K8s, it would require a volume claim, which is not ideal.
3. There are security and availability concerns in this case:
   - We will need to host a download server and make it always available
   - Users may have concerns with downloading code at runtime, as it could be used as an attack vector
   - Some users run Dapr on hermetic environments, where no network connections are allowed

PR #20, opened 2023-04-06
Proposal: Extend service invocation to non-Dapr endpoints 

sicoyle:
Formalizing the discussion for the external service invocation feature here.
Additional context may be found in the [initial issue](https://github.com/dapr/dapr/issues/4549).

### DIFF
--- a/0006-B-external-service-invocation.md
+++ b/0006-B-external-service-invocation.md
@@ -86,13 +86,11 @@ Cons:
 How will this work, technically?
 
 Allow configuration of pieces needed for external service invocation through creation of new CRD titled `ExternalHTTPEndpoint`.
-The sample `yaml` file snippet below represents two examples:
-
-Option 1:
-Option one is HTTP specific in it's `Kind`.
+It is HTTP specific in it's `Kind`.
 This has benefits in being obvious upfront that it supports only `http`,
 and makes it to where we do not need `spec.allowed.protocols`.
 However, it would have the drawback of needing additional CRDs in the future for supporting other protocols such as `gRPC`.
+The sample `yaml` file snippet below represents the proposed configuration.
 
 ```
 apiVersion: dapr.io/v1alpha1
@@ -102,34 +100,6 @@ metadata:
 spec:
   allowed:
   - name: github
-    baseUrl: "github.com"
-    headers:
-    - "Accept-Language": "en-US"
-  metadata:
-  - name: mymetadata
-    secretKeyRef:
-      name: my-secret
-      key: mymetadataSecret
-auth:
-  secretStore: my-secretstore
-```
-
-Option 2:
-Option two is more generic,
-and seamlessly allows for expansion in the future for additional `ExternalEndpoint` feature sets.
-The drawback to this approach is it is less obvious that it only supports `http` for the first iteration,
-but would allow for more flexibility and expansion moving forward.
-This can be seen in the `spec.allowed.protocols` field.
-
-```
-apiVersion: dapr.io/v1alpha1
-kind: ExternalEndpoint
-metadata:
-  name: externalserviceinvocation
-spec:
-  allowed:
-  - name: github
-    protocol: "http"
     baseUrl: "github.com"
     headers:
     - "Accept-Language": "en-US"


0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,169 @@
...
+- Need to know external base URL ahead of time to configure, but that may not always be easy for end users.
+- Would need to change the Dapr Operator to notify on edits for external endpoints.
+
+## Implementation Details
+
+### Design
+
+How will this work, technically?
+
+Allow configuration of pieces needed for external service invocation through creation of new CRD titled `ExternalHTTPEndpointCRD` with following `yaml` specifications:
```
yaron2, line 88: Change to `ExternalHTTPEndpoint`, the `CRD` is redundant

yaron2 (changes requested):

yaron2:
Thanks for the prompt responses to requested changes @sicoyle

+1 binding

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,157 @@
...
+                              properties:
+                                baseUrl:
+                                  type: string
+                                headers:
+                                  items:
+                                    type: string
+                                  type: array
+                                name:
+                                  type: string
+                                protocol:
```
daixiang0, line 105: Is it for future?

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,157 @@
...
+                                  type: string
+                              required:
+                              - baseUrl
+                              - name
```
daixiang0, line 109: What is the name used for? Like UUID? If there is an example, would be better.

daixiang0:
Overall is good, +1

0006-B-external-service-invocation.md
daixiang0: And beside headers, how to deal with the payload?

yaron2: The payload is the same as a regular svc invocation today, it's part of the HTTP call

daixiang0: Thanks.

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,157 @@
...
+        externalServiceInvocation:
+                        properties:
+                          allowed:
+                            items:
+                              description: ExternalAPIAccessSpec describes an access specification for allowing external service invocations.
+                              properties:
+                                baseUrl:
+                                  type: string
+                                headers:
+                                  items:
```
artursouza, line 100: We need to also resolve secrets with secretRef. It can be a follow up in the implementation but would be great to have it here as a planned feature.

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,157 @@
...
+This feature will need to be fully compatible with existing service invocation API.
+
+* Metrics
+Existing service invocation tracing and metrics capabilities when calling external enpoints will be fully functional.
+
+## Completion Checklist
+
+What changes or actions are required to make this proposal complete?
+
+* Code changes
```
artursouza, line 153: Secret resolution here too, please.

artursouza (changes requested):

0006-B-external-service-invocation.md
sicoyle: I added two yaml snippet examples to address this question. On one hand, the `protocol` field would be redundant in the case that we have `kind: ExternalHTTPEndpoint`, but if we did something like `kind: ExternalEndpoint`, then the `spec.allowed.protocol` would be a necessity.

0006-B-external-service-invocation.md
sicoyle: yes, like a unique identifier. Hopefully the example yaml updates in the proposal bring clarity here. If not, I'd be happy to elaborate :)

artursouza:
Please, pick an option. My recommendation is to make it HTTP-specific. We had tried creating a generic layer for HTTP and gRPC but it ends up with odd attributes that are only applicable to one.

sicoyle:
> Please, pick an option. My recommendation is to make it HTTP-specific. We had tried creating a generic layer for HTTP and gRPC but it ends up with odd attributes that are only applicable to one.

I updated the doc selecting the HTTP-specific approach, while dropping the redundant protocols field. Please let me know if anyone has other thoughts on this!

artursouza (approved):

yaron2 (approved):

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,158 @@
...
+The sample `yaml` file snippet below represents the proposed configuration.
+
+```
+apiVersion: dapr.io/v1alpha1
+kind: ExternalHTTPEndpoint
+metadata:
+  name: externalserviceinvocation
+spec:
+  allowed:
+  - name: github
```
mukundansundar, line 102: following from this [comment](https://github.com/dapr/dapr/issues/4549#issuecomment-1152624173) on `name` being used for service discovery, if both internal service invocation and external service invocation define the same `name`, one as the `name` field the other in terms of `app-id`, which will be given priority?

mukundansundar (commented):
was there any specific discussion on how the API will look like and work? Ideally if it is going to be similar to the existing API, could you please add in the API endpoint details. 

For HTTPS endpoints, how are the trust anchors going to be configured, i.e. for internal service invocation mTLS based on sentry is used, what is the case for external invocation?

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,158 @@
...
+they can use an external URL and  optionally overwrite fields at the time of invocation.
+
+
+### Feature lifecycle outline
+
+* Expectations
+This feature will be enabled after Dapr v1.11 release.
+
+* Compatability guarantees
+This feature is fully compatible with the existing service invocation API.
```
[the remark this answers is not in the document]
yaron2: All the existing service invocation URLs will be supported

0006-B-external-service-invocation.md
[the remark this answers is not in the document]
yaron2: name should be unique across the namespace for the external invocation resource

mukundansundar: adding the URL format in this proposal will be helpful.

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,158 @@
...
+apiVersion: dapr.io/v1alpha1
+kind: ExternalHTTPEndpoint
+metadata:
+  name: externalserviceinvocation
+spec:
+  allowed:
+  - name: github
+    baseUrl: "github.com"
+    headers:
+    - "Accept-Language": "en-US"
```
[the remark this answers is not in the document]
yaron2: That should be doable

0006-B-external-service-invocation.md
yaron2: It's the same URL formats as now

ItalyPaleAle:
Sorry I didn't see this sooner.

One comment/question I have is whether it'd be possible to offer a programmatic way to define these external endpoints too, in addition to using a CRD (like we do for pubsub subscriptions for example).

Also,  how would this interact (if it would at all) with mTLS? And how could callers be authenticated?

0006-B-external-service-invocation.md
```diff
@@ -0,0 +1,158 @@
...
+
+User facing changes include overriding the URL when calling Dapr for service invocation.
+Users will use the existing service invocation API, but instead of using an app ID,
+they can use an external URL and  optionally overwrite fields at the time of invocation.
+
+
+### Feature lifecycle outline
+
+* Expectations
+This feature will be enabled after Dapr v1.11 release.
```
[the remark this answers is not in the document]
sicoyle: https://github.com/dapr/proposals/pull/22

ItalyPaleAle:
What do you mean with "the most basic authentication features"?

ItalyPaleAle:
I would argue it is, because there are certain things that are very critical to security:

- Username/password or auth token only are shared secrets so we would need to define how they can be rotated, perhaps with 0-downtime? (I.e. you Can’t restart all apps so they pick up the new secret)
- Should TLS/HTTPS be enforced? Otherwise the secrets are transferred in plain-text.
- Should Dapr generate JWTs to authenticate calls? This is significantly better than shared tokens because they are time- and audience-bound.

yaron2:
In general this is about talking to endpoints that have their own security postures which are out of our control, essentially just like the HTTP binding and components in general which don't necessarily include authentication at all or to begin with.

In this case this is very similar to the HTTP binding and components in general - if the external non-Dapr endpoint has authentication mechanisms that the external invocation doesn't support, then it simply will not be usable. The strictness requirement is on the external endpoint and it's up to Dapr to provide users with the most options to meet that, but that 100% doesn't need to be tackled from the very beginning and can be incremental just like many other authentication mechanisms we have in different places.

Now regardless, here's what I'm thinking about the issues you raised:

 > Username/password or auth token only are shared secrets so we would need to define how they can be rotated, perhaps with 0-downtime? (I.e. you Can’t restart all apps so they pick up the new secret)

Restarting multiple replicas in a rolling fashion can provide zero downtime, and this is exactly what happens today for all of our components when a secret, password or token changes. It might be beneficial to consider a general purpose secret rotation mechanism for Dapr sidecars but this is definitely outside the scope of this proposal

> Should TLS/HTTPS be enforced? Otherwise the secrets are transferred in plain-text

No, the user should use this feature the same as they would an HTTP client talking to an external endpoint. We can provide a behavior that enforces TLS/HTTPS is an authentication secret is provided.

> Should Dapr generate JWTs to authenticate calls? This is significantly better than shared tokens because they are time- and audience-bound

Adding a JWT authentication feature would be good to add in the future.

ItalyPaleAle:
> In general this is about talking to endpoints that have their own security postures which are out of our control, essentially just like the HTTP binding and components in general which don't necessarily include authentication at all or to begin with.

Ok, this is good context-setting. With this premise, I agree that we can push more of the responsibility of authenticating calls towards the user. Let's just make sure this is documented :)

> Restarting multiple replicas in a rolling fashion can provide zero downtime, and this is exactly what happens today for all of our components when a secret, password or token changes. It might be beneficial to consider a general purpose secret rotation mechanism for Dapr sidecars but this is definitely outside the scope of this proposal

The problem is coordination because during the rollout you'll have some apps making calls with the old API token, and some with the new one. However, given the point above, I guess this is not our problem :)

> No, the user should use this feature the same as they would an HTTP client talking to an external endpoint. We can provide a behavior that enforces TLS/HTTPS is an authentication secret is provided.

Given the premise above, I think this is not necessary.

> Adding a JWT authentication feature would be good to add in the future.

Agree. @sicoyle would you mind opening an issue in dapr/dapr so we don't forget about this? :)

lloydfischer:
I’m one of the instigators of the original ask for this feature. The primary use case for us was to allow a DAPR service to call out to an external service without having to get a token. Ideally, middleware would be invoked on the way out that could get a token and add it to the request. Is that possible under this proposal?

yaron2:
> The question is if the call to a non-DAPR endpoint passes through middleware or not

Yes, these calls will pass through the middleware.

PR #21, opened 2023-04-07
Proposal: Dapr Error Handling Codes

robertojrojas:
This is a proposal for Dapr Error Handling and Codes standarization.
Issue: https://github.com/dapr/dapr/issues/6068


000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,337 @@
...
+|404  |	NOT_FOUND |	A specified resource is not found. |
+|409  |	ABORTED	| Concurrency conflict, such as read-modify-write conflict. |
+
+### Error Details Classification
+The following tables shows the propsosed error codes used in the **reason** for the **google.rpc.ErrorInfo** within the various levels of Dapr modules:
+
+**INIT**
+| Dapr Module | Description |
+| ----------- | ----------- |
+| CLI         | DPR_INIT_CLI_[1-1000] |
```
artursouza, line 138: nit: just spell out DAPR since it is not saving much.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,337 @@
...
+| Self-hosted | DPR_INIT_SLF_[1-1000] |
```
artursouza, line 139: Unless we really need the space, can we just spell it out? If we need to save space, we can truncate the message.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,337 @@
...
+### Sample Error
+```json
+{
+  "code": 8089,
+  "message": "unable to perform Redis.HGETALL call: dial tcp [::1]:6379: connect: connection refused",
+  "details": [
+    {
+      "@type": "type.googleapis.com/google.rpc.ErrorInfo",
+      "reason": "DPR_COMP_STS_1",
+      "domain": "redis.state.contrib.dapr.io",
```
artursouza, line 174: I think the domain should not contain the component type. It feels that there is enough information in the reason that would make the domain simply be "dapr.io" - see how reason has "STS" and domain has "state", reason has "COMP" and domain has "contrib".

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,337 @@
...
+      "metadata": {
+        "key": "myapp||name"
+      }
+    },
+    {
+      "@type": "type.googleapis.com/google.rpc.ResourceInfo",
+      "resource_type": "StateStore",
+      "resource_name": "REDIS",
```
artursouza, line 182: This is good, it means we can keep ErrorInfo agnostic of component implementation.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,337 @@
...
+}
+```
+### Pros
+- Since the Dapr Runtime is using protocol buffers as the data format, support for the richer error model is already included in most of the gRPC implementations.
+- This would help minimize the changes with the Dapr ecosystem.
+- This solution could be used to programmatically react to errors as it provides a standard structure for the errors with details.
+
+### Cons
+- Dependencies on gPRC richer error model.
+- Not all gRPC implementations support this out of the box, [C#/dotnet](https://learn.microsoft.com/en-us/dotnet/architecture/grpc-for-wcf-developers/error-handling#grpc-richer-error-model) being one of them, but there is a [possible solution (Google.Api.CommonProtos)](https://www.nuget.org/packages/Google.Api.CommonProtos/) to this.
```
artursouza, line 265: This is a big deal, but we need to confirm with @halspang if the extra dependency is OK.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,337 @@
...
+
+- client/state.go
+
+https://github.com/robertojrojas/dapr-cli/tree/error-codes-poc
+
+- pkg/standalone/invoke.go
+
+
+
+## Solution B
```
artursouza, line 314: Mention this as alternative consideration but the proposal should have only one solution and not two.

artursouza (changes requested):

000X-B-C-I-R-S-error-handling-codes.md
robertojrojas: `dotnet-sdk` works well. I need to test `js-sdk`, `java-sdk`, and `python-sdk`

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,319 @@
...
+
+### Sample Error
+```json
+{
+  "code": 8089,
+  "message": "unable to perform Redis.HGETALL call: dial tcp [::1]:6379: connect: connection refused",
+  "details": [
+    {
+      "@type": "type.googleapis.com/google.rpc.ErrorInfo",
+      "reason": "DAPR_STATE_STORE_COMPONENT",
```
artursouza, line 173: The reason should be more specific for the client to make use of it. For example, was it a eTag mismatch or a random failure trying to write?

artursouza: DAPR_STATE_ETAG_MISMATCH

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,319 @@
...
+| CLI         | DAPR_CLI_RUNTIME |
+| Self-hosted | DAPR_SELF_HOSTED_RUNTIME |
+| dapr-2-dapr(HTTP) | DAPR_HTTP_RUNTIME |
+| dapr-2-dapr(gRPC) | DAPR_GRPC_RUNTIME |
+
+**COMPONENTS**
+| Dapr Module | Description |
+| ----------- | ----------- |
+| PubSub              | DAPR_PUBSUB_COMPONENT |
+| StateStore          | DAPR_STATE_STORE_COMPONENT |
```
artursouza, line 155: Same here, DAPR_STATE_*

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,319 @@
...
+|404  |	NOT_FOUND |	A specified resource is not found. |
+|409  |	ABORTED	| Concurrency conflict, such as read-modify-write conflict. |
+
+### Error Details Classification
+The following tables shows the propsosed error codes used in the **reason** for the **google.rpc.ErrorInfo** within the various levels of Dapr modules:
+
+**INIT**
+| Dapr Module | Description |
+| ----------- | ----------- |
+| CLI         | DAPR_CLI_INIT |
```
artursouza, line 138: Prefix only: DAPR_CLI_*

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+| ----------- | ----------- |
+| CLI         | DAPR_CLI_INIT_* |
+| Self-hosted | DAPR_SELF_HOSTED_INIT_* |
+| K8S         | DAPR_K8S_INIT_* |
+| Invoke      | DAPR_INVOKE_INIT_* |
+
+**RUNTIME**
+| Dapr Module | Prefix |
+| ----------- | ----------- |
+| CLI         | DAPR_CLI_RUNTIME_* |
```
mukundansundar, line 168: what exactly does CLI RUNTIME mean here? 
Probably just `DAPR_RUNTIME_*`? 
Or Probably just `DAPR_EXEC_*`?

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+
+
+### Error Details Prefixes
+The following tables shows the propsosed error codes prefixes used in the **reason** for the **google.rpc.ErrorInfo** for various Dapr building blocks:
+
+
+**INIT**
+| Dapr Module | Prefix |
+| ----------- | ----------- |
+| CLI         | DAPR_CLI_INIT_* |
```
mukundansundar, line 160: Can we leave it at `DAPR_INIT_*`? We do have a separate dapr `CLI` and `daprd` is a binary and not a CLI.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+error structure. This is defined in the [error_details.proto Protobuf](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) 
+
+```go
+message ErrorInfo {
+  // The reason of the error. This is a constant value that identifies the
+  // proximate cause of the error. Error reasons are unique within a particular
+  // domain of errors. This should be at most 63 characters and match a
+  // regular expression of `[A-Z][A-Z0-9_]+[A-Z0-9]`, which represents
+  // UPPER_SNAKE_CASE.
+  string reason = 1;
```
mukundansundar, line 71: is there a particular max length that is recommended for this?

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+
+
+Below is partial table of the Standard Error code provided by gRPC and how they map to HTTP error codes. The entire list can found in the following links:
+- [Google API Error Handling]https://cloud.google.com/apis/design/errors#handling_errors
+- (gRPC Codes ProtoBuf)[https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto]
+
+
+|HTTP | gRPC      | Description |
+|---- | ----------- | ----------- |
+|200  |	OK		  | No error. |
```
mukundansundar, line 114: Probably list all the current set of HTTP return codes and their mappings that dapr currently returns.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Error Details Prefixes
+The following tables shows the propsosed error codes prefixes used in the **reason** for the **google.rpc.ErrorInfo** for various Dapr building blocks:
+
+
+**INIT**
+| Dapr Module | Prefix |
+| ----------- | ----------- |
+| CLI         | DAPR_CLI_INIT_* |
+| Self-hosted | DAPR_SELF_HOSTED_INIT_* |
```
mukundansundar, line 161: I think, these should be completely platform agnostic.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+| K8S         | DAPR_K8S_INIT_* |
+| Invoke      | DAPR_INVOKE_INIT_* |
+
+**RUNTIME**
+| Dapr Module | Prefix |
+| ----------- | ----------- |
+| CLI         | DAPR_CLI_RUNTIME_* |
+| Self-hosted | DAPR_SELF_HOSTED_RUNTIME_* |
```
mukundansundar, line 169: remove platform specific codes.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+| dapr-2-dapr(HTTP) | DAPR_HTTP_RUNTIME_* |
+| dapr-2-dapr(gRPC) | DAPR_GRPC_RUNTIME_* |
```
mukundansundar, line 171: the other way around 
```
DAPR_RUNTIME_HTTP_*
DAPR_RUNTIME_GRPC_*
```

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,307 @@
...
+  // UPPER_SNAKE_CASE.
+  string reason = 1;
+
+  // The logical grouping to which the "reason" belongs. The error domain
+  // is typically the registered service name of the tool or product that
+  // generates the error. Example: "pubsub.googleapis.com". If the error is
+  // generated by some common infrastructure, the error domain must be a
+  // globally unique value that identifies the infrastructure. For Google API
+  // infrastructure, the error domain is "googleapis.com".
+  string domain = 2;
```
mukundansundar, line 79: is it going to be only `dapr.io` here? And if so do we need the domain field?

mukundansundar (commented):
first pass review.

000X-B-C-I-R-S-error-handling-codes.md
robertojrojas: `CLI RUNTIME` would mean something like `INVOKE`, etc... As we dig deeper into each building block we can redefine some of these.

000X-B-C-I-R-S-error-handling-codes.md
robertojrojas: As we dig deeper into each building block we can redefine some of these.

000X-B-C-I-R-S-error-handling-codes.md
robertojrojas: 63 characters appears to be the max length. I haven't found out why, but to be on the safe side we should stick to it.

000X-B-C-I-R-S-error-handling-codes.md
robertojrojas: At the moment, we don't have a mapping to Dapr specific errors. As we discussed in the review call, we should know more as we dig deeper.

000X-B-C-I-R-S-error-handling-codes.md
robertojrojas: For now we're planning it to be `dapr.io`.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,309 @@
...
+
+
+|HTTP | gRPC      | Description |
+|---- | ----------- | ----------- |
+|200  |	OK		  | No error. |
+|400  |	INVALID_ARGUMENT |	Client specified an invalid argument. Check error message and error details for more information. |
+|400  |	FAILED_PRECONDITION |	Request can not be executed in the current system state, such as deleting a non-empty directory. |
+|400  |	OUT_OF_RANGE |	Client specified an invalid range. |
+|401  |	UNAUTHENTICATED	| Request not authenticated due to missing, invalid, or expired OAuth token. |
+|403  |	PERMISSION_DENIED |	Client does not have sufficient permission. This can happen because the OAuth token does not have the right scopes, the client doesn't have permission, or the API has not been enabled. |
```
ItalyPaleAle, line 119: Nit: we don't use OAuth.

```suggestion
|401  |	UNAUTHENTICATED	| Request not authenticated due to missing, invalid, or expired authorization credentials. |
|403  |	PERMISSION_DENIED |	Client does not have sufficient permission. |
```

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,309 @@
...
+      "@type": "type.googleapis.com/google.rpc.ErrorInfo",
+      "reason": "DAPR_STATE_ETAG_MISMATCH",
+      "domain": "dapr.io",
+      "metadata": {
+        "key": "myapp||name"
+      }
+    },
+    {
+      "@type": "type.googleapis.com/google.rpc.ResourceInfo",
+      "resource_type": "StateStore",
```
ItalyPaleAle, line 203: Some keys use camelCase, some use snake_case. We should be consistent

Also see my comment above. I think a single `resource_name` = `state.redis/v1` may be more aligned with how we refer to the comoponents everywhere (note that some components have a dot in the name too, such as `state.azure.cosmosdb`, and we should include the version too)

ItalyPaleAle (commented):
Would love to see an example of how this is included in a HTTP response too.

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,309 @@
...
+  // {"instanceLimitPerRequest": "100"}, if the client exceeds the number of
+  // instances that can be created in a single (batch) request.
+  map<string, string> metadata = 3;
+}
+```
+
+### Error Status
+The properties of the **google.rpc.Status** will be populated as following:
+
+- **Code**  - Protocol level error code. These could be either gRPC or HTTP error codes. See (gRPC Codes ProtoBuf)[https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto]
```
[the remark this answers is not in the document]
robertojrojas: We could address the HTTP error codes with a mapping like the one shown in the table: https://cloud.google.com/apis/design/errors#handling_errors

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,309 @@
...
+### ResourceInfo (Optional)
+The properties of the **type.googleapis.com/google.rpc.ResourceInfo** will be populated as following:
+
+- **ResourceType** - The building block type.
+   
+   Example: "StateStore"
+
+- **ResourceName** - The component name.
+   
+   Example: "REDIS"
```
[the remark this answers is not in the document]
robertojrojas: I like `state.redis/v1` for `ResourceName`. 
@artursouza any thoughts?

000X-B-C-I-R-S-error-handling-codes.md
```diff
@@ -0,0 +1,309 @@
...
+| Self-hosted | DAPR_SELF_HOSTED_INIT_* |
+| K8S         | DAPR_K8S_INIT_* |
+| Invoke      | DAPR_INVOKE_INIT_* |
+
+**RUNTIME**
+| Dapr Module | Prefix |
+| ----------- | ----------- |
+| CLI         | DAPR_CLI_RUNTIME_* |
+| Self-hosted | DAPR_SELF_HOSTED_RUNTIME_* |
+| dapr-2-dapr(HTTP) | DAPR_HTTP_RUNTIME_* |
```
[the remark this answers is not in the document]
robertojrojas: @ItalyPaleAle This is one of those part of Darp I'm not totally clear. I added the HTTP protocol for Dapr-2-Dapr thinking that Dapr side cars have an option of replacing gRPC for HTTP for internal communication of each other. isn't that the case?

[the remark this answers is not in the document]
ItalyPaleAle: Internal communication (sidecar-to-sidecar) always happens over gRPC.

000X-B-C-I-R-S-error-handling-codes.md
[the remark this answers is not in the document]
ItalyPaleAle: How are errors returned in HTTP, in general?

000X-B-C-I-R-S-error-handling-codes.md
robertojrojas: The `keys`  for the JSON payload are auto-generated as part of the Marshaling implemented by the Protobuf stub of the [error_details.proto](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto) 

if you look at the annotations, you'll see the `resource_type` as the JSON key to generate/read.
```
ResourceType string `protobuf:"bytes,1,opt,name=resource_type,json=resourceType,proto3" json:"resource_type,omitempty"`
```

000X-B-C-I-R-S-error-handling-codes.md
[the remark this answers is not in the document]
robertojrojas: @ItalyPaleAle  Here what I did as part of my POC, but of course it could implemented better: https://github.com/robertojrojas/dapr/blob/a0760ea549b9a3da93af2bee96e88ad94db9b32b/pkg/http/api.go#L1017

000X-B-C-I-R-S-error-handling-codes.md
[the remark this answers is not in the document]
ItalyPaleAle: Great callout.

Either a different key or `state.redis/v1/component-name`?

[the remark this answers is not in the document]
artursouza: I am fine with `state.redis/v1/component-name`

000X-B-C-I-R-S-error-handling-codes.md
artursouza: ErrorInfo is not out structure, it is Google's gRPC and we will pass `dapr.io`, so SDKs can know which ones they care about in the array of ErrorInfo.

artursouza:
I made some changes that reflect the latest discussions. Can we merge this? @mukundansundar @ItalyPaleAle

yaron2 (approved):
+1 binding

artursouza (approved):
+1 binding

daixiang0 (approved):
+1 binding

PR #33, opened 2023-07-14
Sdk resiliency

artursouza:
Proposal for https://github.com/dapr/dapr/issues/6609

### DIFF
--- a/0009-S-sdk-resiliency.md
+++ b/0009-S-sdk-resiliency.md
@@ -66,11 +66,16 @@ Cons:
 ### Design
 
 * `DAPR_API_MAX_RETRIES` defines the maximum number of retries, SDKs can determine which strategy will be implemented (linear, exponential backoff, etc). `0` is the default value and means no retry (same for negative values).
-* `DAPR_API_TIMEOUT_SECONDS` defines the maximum waiting time to connect and receive a response for an HTTP or gRPC call. Defaults to `10`. `0` (or negative) are invalid values and it will default to `10` - warning to be printed in that case. This setting is the timeout for each API invocation and not the timeout of the aggregated time for retries. This setting can be used without retries.
+* `DAPR_API_TIMEOUT_SECONDS` defines the maximum waiting time to connect and receive a response for an HTTP or gRPC call. Defaults to `0`. `0` (or negative) are handled as "undefined" and calls might hang forever on the client side. This setting is the timeout for each API invocation and not the timeout of the aggregated time for retries. This setting can be used without retries.
+* All environment variables can be overwritten via parameters to the Dapr client or at a per-request basis, in the following order (higher priority on top):
+  1. Per-request parameter
+  2. Parameter when instantiating a Dapr client object
+  3. Properties or any other language specific configuration framework.
+  4. Environment variables
 * SDK to retry if error is on connection.
 * SDK to retry in case of the following retriable codes:
   * gRPC: DEADLINE_EXCEEDED, UNAVAILABLE.
-  * HTTP: 408, 429, 500, 502, 503, 504
+  * HTTP: 408, 429 (respect `Retry-After` header), 500, 502, 503, 504
 * The same client should still be usable if the API goes down but is restored after any arbitrary amount of time. In other words, the unavailability of the Dapr API should not require the application to restart.
 
 #### Example of implementation
@@ -80,13 +85,13 @@ https://github.com/dapr/java-sdk/pull/889
 ### Feature lifecycle outline
 
 * Compatability guarantees
-Retries should be disabled by default. Timeouts, on the other hand, should be enforced with the default value.
+Retries and timeouts should be disabled by default.
 
 * Deprecation / co-existence with existing functionality
 If customers prefer to have a more fine tuned resiliency logic, they can still achieve so by disabling the SDK resiliency and use a 3rd party library to handle retries with custom logic.
 
 * Feature flags
-Retries are disabled by default with the value `0`.
+Retries and timeouts are disabled by default with the value `0`.
 
 ### Acceptance Criteria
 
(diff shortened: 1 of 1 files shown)


jjcollinge:
Just sharing this proposal from upstream gRPC on built in retry policies as it is implement in some languages and shows what you can expect from the gRPC clients. https://github.com/grpc/proposal/blob/master/A6-client-retries.md#retryable-status-codes

Also, handling of 429 should ideally respect a `Retry-After` header or the user should be able to opt out of retrying on that status code.

0009-S-sdk-resiliency.md
```diff
@@ -0,0 +1,115 @@
...
+
+Cons:
+- See trade-offs mentiond above.
+
+## Implementation Details
+
+### Design
+
+* `DAPR_API_MAX_RETRIES` defines the maximum number of retries, SDKs can determine which strategy will be implemented (linear, exponential backoff, etc). `0` is the default value and means no retry (same for negative values).
+* `DAPR_API_TIMEOUT_SECONDS` defines the maximum waiting time to connect and receive a response for an HTTP or gRPC call. Defaults to `10`. `0` (or negative) are invalid values and it will default to `10` - warning to be printed in that case. This setting is the timeout for each API invocation and not the timeout of the aggregated time for retries. This setting can be used without retries.
```
ItalyPaleAle, line 69: I think the timeout needs to be defined (or definable) per-request. For example:

- Requests that save state in a state store or perform service invocation can take a long time if there's lots of data
- Requests that are long-lived streams should not have a timeout (we currently have this in gRPC but we will eventually have them in HTTP/2 handlers too)

Also connected to the point above, that making this an env vars can introduce subtle bugs due to the app running in a different enviornment (like "works on my machine").

ItalyPaleAle: Also, the current default is "0" so making this "10" will be a breaking change

JoshVanL: > Requests that are long-lived streams should not have a timeout (we currently have this in gRPC but we will eventually have them in HTTP/2 handlers too)

@ItalyPaleAle why should streams not have a timeout?

ItalyPaleAle: They’re long-lived by design. For example, the Configuration API has a stream that is used to send notifications of updated configurations. This should never time out.

JoshVanL: I agree, but we should still have a timeout on the initial connection dial imo

0009-S-sdk-resiliency.md
```diff
@@ -0,0 +1,115 @@
...
+- Document expected behavior for SDKs regarding timeout and retries
+
+Cons:
+- See trade-offs mentiond above.
+
+## Implementation Details
+
+### Design
+
+* `DAPR_API_MAX_RETRIES` defines the maximum number of retries, SDKs can determine which strategy will be implemented (linear, exponential backoff, etc). `0` is the default value and means no retry (same for negative values).
```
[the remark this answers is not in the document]
artursouza: It depends, Dapr implemented resiliency outside code. We can offer both ways here.

artursouza:
> Just sharing this RFC from upstream gRPC on built in retry policies as it is implement in some languages and shows what you can expect from the gRPC clients. https://github.com/grpc/proposal/blob/master/A6-client-retries.md#retryable-status-codes
> 
> Also, handling of 429 should ideally respect a `Retry-After` header or the user should be able to opt out of retrying on that status code.
> 
> If I read this correctly it's up to the SDK to decide on the retry strategy? If the goal is to provide consistency here - why not make this consistent or configurable?

Regarding 429, I think SDKs should respect the header. I will mention that. We can make this more configurable too, but I am concerned about it exploding into a huge resiliency spec, like in runtime.

Regarding the resiliency strategy, I did not add it here to reduce the scope of the discussion. We need something today and we can expand into more fine grained details later. My concern, again, is to have this to expand into a full resiliency spec like we have in runtime and this taking far too long to be agreed and implemented. We need something today since the experience is poor already if the API shows any connectivity issues.

0009-S-sdk-resiliency.md
ItalyPaleAle: The last changes have addressed my feedback and concerns.

artursouza:
I have updated the proposal. Changes:

1. Timeout defaults to "undefined", so it is not a breaking change.
2. Add params per-request and per-client to overwrite the ENV variable when needed (including priority among them)
3. Respect Retry-After header for 429

I have not detailed about how the retry is implemented. Maybe maintainers from SDK have also opinions about this.

0009-S-sdk-resiliency.md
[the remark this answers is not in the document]
artursouza: Done.

msfussell:
Updated proposal looks good.

JoshVanL:
+1

yaron2:
+1 binding

artursouza:
+1 binding

shubham1172:
+1 binding

XavierGeerinck:
+1! definitely helpful as it would resolve issues I am having with running on App Service :) (requirement sadly enough)

0009-S-sdk-resiliency.md
```diff
@@ -0,0 +1,122 @@
...
+### Feature lifecycle outline
+
+* Compatability guarantees
+Retries and timeouts should be disabled by default.
+
+* Deprecation / co-existence with existing functionality
+If customers prefer to have a more fine tuned resiliency logic, they can still achieve so by disabling the SDK resiliency and use a 3rd party library to handle retries with custom logic.
+
+* Feature flags
+Retries and timeouts are disabled by default with the value `0`.
```
halspang, line 94: Is it also worth calling out the infinite retry case here? Do we want to support it? Will it be a `-1` like runtime? Will it be a special policy?

artursouza: I am not adding infinite retries as a thing now. It does not mean it cannot be added later. Thoughts?

artursouza: Let me set it as -1 to infinity in this proposal to avoid a breaking change in the future. Default is still 0 for SDKs.

halspang:
+1 binding

artursouza:
-1 now means infinity for retries - same behavior as in runtime.

yaron2:
Proposal is approved.

yaron2 (approved):

PR #34, opened 2023-07-20
Add component healthcheck api design

DeepanshuA:
API Design, Approaches and Recommended Approach for Components Healthcheck in Dapr

### DIFF
--- a/0010-R-components-healthcheck.md
+++ b/0010-R-components-healthcheck.md
@@ -1,0 +1,477 @@
+# Components Healthcheck In Dapr
+
+* Author(s): Deepanshu Agarwal (@DeepanshuA)
+* State: Draft
+* Updated: 07/20/2023
+
+## Overview
+As a user, I should be able to enquire the health of dapr system, which is possible currently - but this health check should also include health of connected dapr components as well.
+
+## Tenets
+1. It should be extensible i.e. tomorrow if any feature to check health of Actors etc. is required, it should not require a new endpoint.
+2. It should not give any false positives or false negatives.
+
+## Current Scenario
+There are many components in Dapr which don't yet implement Ping. 
+Earlier, Ping was made mandatory due to which it used to give false positive in various components. But, then it was made optional and only those components have Ping function, which properly implement it.
+So, how to provide a correct answer to users, if we can't enquire all components?
+
+
+## Moving Ahead
+If we try to give a one-word Yes/No kind of answer to user for the Question: "Are all components healthy in my system?", that means that if he/she has some component(s), which don't yet implment Ping, then it can't be a single word answer "Yes" or "No". We definitely need to provide him/her the information that what all components couldn't be checked through this check, as they don't yet implement Ping function.
+We should definitely have some Work Items to ensure that most Components start implementing Ping, but that doesn't mean that we can't go ahead with Components health check for now, we just need to ensure that we are providing a proper info about those components explicitly to user which are in his system but could not be verified via the health check.
+
+
+## API Design
+
+Most of the Approaches underneath work with a query parameter, in addition to `healthz` endpoint. Instead of, an additional endpoint.
+
+Following are the possible responses for `healthz` API:
+
+| HTTP | Response Codes | 
+| -------- | -------- | 
+| 204     | dapr is healthy     | 
+| 500     | dapr is not healthy     | 
+
+Hence, if dapr is healthy, then the response code changes, as per the enlisted case below.
+If dapr is NOT healthy, then the compoenents Health check should anyways NOT be checked.
+
+
+### Approach 1:
+List only failed or tose componnets for which Ping is not implemented. Create separate sections for "Failed" and "Ping Not Implemented".
+
+HTTP Endpoint: http://localhost:3500/v1.0/healthz?components=true
+GRPC Endpoint: GetComponentHealthAlpha1 (It will be separate endpoint for grpc)
+
+**Case 1:** When ALL components in system implement Ping and are Healthy
+
+Result:
+
+Response code: 204 OK No Content
+GRPC Response code (if implemented): 0 OK
+
+**Case 2:** When ALL components in system implement Ping BUT some components have failed Ping check: Here we report only failed components and not passed components, we don't treat this API as a way to list components.
+
+HTTP Response code: 500 Internal Server Error
+GRPC Response code (if implemented): 2 UNKNOWN
+
+```
+{
+    "componentsHealth": 
+        "failedComponents": [
+            {
+                "componentName": "txnstore",
+                "type": "state",
+                "status": "NOT_OK",
+                "message": "redis store: error connecting to redis at localhost:6379: dial tcp 127.0.0.1:6379: connect: connection refused"
+            },
+            {
+                "componentName": "smsbinding",
+                "type": "bindings",
+                "status": "NOT OK",
+                "message": "redis binding: error connecting to redis at localhost:6379: dial tcp 127.0.0.1:6379: connect: connection refused"
+            }
+        ]
+}
+```
+**Case 3:** When SOME components in system implement Ping AND some components DON'T implement Ping, AND some components have failed Ping check as well: Here we report failed components and those components as well for which Ping is not implemented, we don't treat this API as a way to list components.
+
+HTTP Response code: 500 Internal Server Error
+GRPC Response code (if implemented): 2 UNKNOWN
+
+```
+{
+    "componentsHealth": 
+        "failedComponents": [
+            {
+                "componentName": "txnstore",
+                "type": "state",
+                "status": "NOT_OK",
+                "message": "redis store: error connecting to redis at localhost:6379: dial tcp 127.0.0.1:6379: connect: connection refused"
+            },
+            {
+                "componentName": "smsbinding",
+                "type": "bindings",
+                "status": "NOT OK",
+                "message": "redis binding: error connecting to redis at localhost:6379: dial tcp 127.0.0.1:6379: connect: connection refused"
+            }
+        ],
+        "pingNotImplemented": [
+            {
+                "componentName": "mockABC",
+                "type": "pubsub",
+            },
+            {
+                "componentName": "mockXYX",
+                "type": "bindings",
+            }
+        ]
+}
+```
+
+**Case 4:** When ANY component in system DOES NOT implement Ping: Here we report ALL those components for which Ping is not implemented.
+
+HTTP Response code: 405 Method Not Allowed
+GRPC Response code (if implemented): 12 UNIMPLEMENTED
+
+```
+{
+    "componentsHealth": 
+        "pingNotImplemented": [
+            {
+                "componentName": "txnstore",
+                "type": "state",
+            },
+            {
+                "componentName": "smsbinding",
+                "type": "bindings",
+            },
+            {
+                "componentName": "mockABC",
+                "type": "pubsub",
+            },
+            {
+                "componentName": "mockXYX",
+                "type": "bindings",
+            }
+        ]
+}
+```
+**Case 5:** When SOME components in system implement Ping and whatever Compoenents implement Ping are Healthy: Here we report only those components for which Ping is not implemented.
+
+HTTP Response code: 200 OK
+GRPC Response code (if implemented): 0 OK
+
+```
+{
+    "componentsHealth": 
+        "pingNotImplemented": [
+            {
+                "componentName": "mockABC",
+                "type": "pubsub",
+            },
+            {
+                "componentName": "mockXYX",
+                "type": "bindings",
+            }
+        ]
+}
+```
+
+**Pros of this Approach:**
+1. Only required components (failed or Ping not implemented) are listed in result.
+
+
+**Cons of this Approach:**
+1. Different umbrellas for "Failed Components" and "Ping Not Implemented" - doesn't give a consistent view and may require very custom parsing logic, if users have some downstream system to listen to this heartbeat.
+
+### Approach 2:
+Instead of listing under umbrellas, like "Failed components" or "Ping Not Implemented", list all components and report accordingly for them.
+
+HTTP Endpoint: http://localhost:3500/v1.0/healthz?components=true
+GRPC Endpoint: GetComponentHealthAlpha1 (It will be separate endpoint for grpc)
+**Case 1:** When ALL components in system implement Ping and are Healthy
+
+Result:
+
+Response code: 204 OK No Content
+GRPC Response code (if implemented): 0 OK
+```
+{
+    "componentsHealth": 
+        {
+            "componentName": "txnstore",
+            "type": "state",
+            "status": "OK",
+        },
+        {
+            "componentName": "smsbinding",
+            "type": "bindings",
+            "status": "OK",
+        }
+}
+```
+
+**Case 2:** When ALL components in system implement Ping BUT some/all components have failed Ping check:
+
+HTTP Response code: 500 Internal Server Error
+GRPC Response code (if implemented): 2 UNKNOWN
+
+```
+{
+    "componentsHealth": 
+        {
+            "componentName": "txnstore",
+            "type": "state",
+            "status": "NOT_OK",
+            "message": "redis store: error connecting to redis at localhost:6379: dial tcp 127.0.0.1:6379: connect: connection refused"
+        },
+        {
+            "componentName": "smsbinding",
+            "type": "bindings",
+            "status": "NOT OK",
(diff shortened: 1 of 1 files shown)


0010-R-components-healthcheck.md
```diff
@@ -0,0 +1,477 @@
...
+
+## Overview
+As a user, I should be able to enquire the health of dapr system, which is possible currently - but this health check should also include health of connected dapr components as well.
+
+## Tenets
+1. It should be extensible i.e. tomorrow if any feature to check health of Actors etc. is required, it should not require a new endpoint.
+2. It should not give any false positives or false negatives.
+
+## Current Scenario
+There are many components in Dapr which don't yet implement Ping.
```
DeepanshuA, line 15: Comment from Alessandro (@ItalyPaleAle ): I don't know if making Ping mandatory is needed. A lot of components are stateless (for example, they don't maintain persistent connections with a remove service). IMHO it's fine to include Ping in an optional interface.

Ref: https://hackmd.io/MaNpUYRyQqe-0eqCsSUiIg

DeepanshuA: Yeah, it is Optional only right now. And, that is the correct state in my opinion too. The doc also doesn't recommend it to make mandatory.

DeepanshuA, line 15: Comment from Alessandro (@ItalyPaleAle ): That said, we should review components that don't implement Ping, and see if adding it would be useful.

Ref: https://hackmd.io/MaNpUYRyQqe-0eqCsSUiIg

DeepanshuA: Agreed

0010-R-components-healthcheck.md
```diff
@@ -0,0 +1,477 @@
...
+And, to implement only http endpoint.
+
+### All Approaches with Question:
+Should we support query for a partular component Health check?
+I don't think that is required specifically, as of now.
+It is more about overall health of the system.
+
+
+## Open Questions
+1. Is it only for kubernetes users? Is it only needed for http endpoint? Or we should cover gRPC endpoint as well?
```
DeepanshuA, line 476: Comment from Alessandro (@ItalyPaleAle ): No, although K8s is arguably the main user of this.

Ref: https://hackmd.io/MaNpUYRyQqe-0eqCsSUiIg

DeepanshuA: Comment from Alessandro (@ItalyPaleAle ): Which brings up a question: should we add an annotation to indicate which components to include in the K8s healthchecks?

Ref: https://hackmd.io/MaNpUYRyQqe-0eqCsSUiIg

DeepanshuA, line 476: Comment from Alessandro (@ItalyPaleAle ): We don't have a gRPC healthcheck endpoint today. Not sure if that's needed?

Ref: https://hackmd.io/MaNpUYRyQqe-0eqCsSUiIg

0010-R-components-healthcheck.md
```diff
@@ -0,0 +1,477 @@
...
+            {
+                "componentName": "smsbinding",
+                "type": "bindings",
+                "status": "NOT OK",
+                "message": "redis binding: error connecting to redis at localhost:6379: dial tcp 127.0.0.1:6379: connect: connection refused"
+            }
+        ]
+}
+```
+**Case 3:** When SOME components in system implement Ping AND some components DON'T implement Ping, AND some components have failed Ping check as well: Here we report failed components and those components as well for which Ping is not implemented, we don't treat this API as a way to list components.
```
DeepanshuA, line 77: Comment from Alessandro (@ItalyPaleAle ): I am not sure we need a case for this. As per my other comment, IMHO it's totally fine for a component to NOT have Ping implemented. I think the healthcheck should invoke Ping when available, and otherwise just skip the component

Ref: https://hackmd.io/MaNpUYRyQqe-0eqCsSUiIg

DeepanshuA: But, won't it give False Positives then? How as a user am I supposed to know that what all components implement Ping and  that only those components are covered by this Health Check?

0010-R-components-healthcheck.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Approach: 
+- Maintain a cache with status of all components loaded successfully and keep updating this cache in a background go routine at a configurable `pingUpdateFrequency`. By default, `pingUpdateFrequency` to be 5 minutes.
+
+- This cache will not start to be built, right at the boot of daprd sidecar. There will be flag (let's say `collectPings`), which will be `false` at the beginning of the daprd sidecar and which will be turned `true`, once all the components are ready.
+Once, `collectPings` is `true`, the cache will start to be populated.
+
+- But, what happens if a component fails to initialize? 
+If a mandatory component fails to initialize, then daprd will not come up healthy and thus, anyways health check will report unhealthy.
+If an optional component fails to initialize OR if it is not healthy afterwards as well, then it is governed by a query parameter `ignoreOptionalComponent`, which is `true` by default. So, as the name suggests, if this query param is not set to `false`, then optional components failure to initialize OR failure to report healthy will not result in a un-healthy status. Rather, the http status code 207 will be reported back.
```
ItalyPaleAle, line 42: `ignoreOptionalComponent` uses camelCase, while `include_components` uses snake_case. Let's be consistent here please

0010-R-components-healthcheck.md
```diff
@@ -0,0 +1,51 @@
...
+
+Hence, if dapr is healthy, then the response code changes, as per the enlisted cases below.
+If dapr is NOT healthy, then the compoenents Health check should anyways NOT be checked.
+
+http://localhost:3500/v1.0/healthz?include_components=true
+
+### Approach: 
+- Maintain a cache with status of all components loaded successfully and keep updating this cache in a background go routine at a configurable `pingUpdateFrequency`. By default, `pingUpdateFrequency` to be 5 minutes.
+
+- This cache will not start to be built, right at the boot of daprd sidecar. There will be flag (let's say `collectPings`), which will be `false` at the beginning of the daprd sidecar and which will be turned `true`, once all the components are ready.
```
ItalyPaleAle, line 37: How are these options passed? Are they in the Configuration CRD?

0010-R-components-healthcheck.md
```diff
@@ -0,0 +1,51 @@
...
+| 204     | dapr is healthy     | 
+| 500     | dapr is not healthy     | 
+
+Hence, if dapr is healthy, then the response code changes, as per the enlisted cases below.
+If dapr is NOT healthy, then the compoenents Health check should anyways NOT be checked.
+
+http://localhost:3500/v1.0/healthz?include_components=true
+
+### Approach: 
+- Maintain a cache with status of all components loaded successfully and keep updating this cache in a background go routine at a configurable `pingUpdateFrequency`. By default, `pingUpdateFrequency` to be 5 minutes.
```
ItalyPaleAle, line 35: Some thoughts:

- 5 mins seems a lot? That would mean the time to detect a failure can be as high as 5 mins. Maybe that's just my opinion.
- Ping updates may be more frequent if the component is un-healthy, since we want to see if we can recover faster

olitomlinson:
Before I put too much thought into the API design, I want to get the purpose straight in my head first....

I don't know if this is an over simplification or I've just got this plain wrong, but in my head, I had `component health checks` providing the same outcome as `app health checks`?

i.e. Read this [doc](https://docs.dapr.io/developing-applications/building-blocks/observability/app-health/) on `App Health checks`, and in your mind, substitute the word `app` / `application` for `component` -- Is that what the component health checks are trying to achieve ?

If not, what are the key differences in how the sidecar would behave?

---

At the point of authoring user-code which depends on dapr, I would generally only care about the overall health of the sidecar, which would include the health of all the components too when determining the overall health.

If just one of the components failed their health check, I would expect the the overall health check to fail.

Although I care about the health of the components, I'm unlikely to need to know exactly which components are unhealthy at this point in my code. I'm not saying that the breakdown of component health information isn't useful in other use-cases, but I'm just not sure I would utilise that information in my user-code.

```c#
var client = new DaprClientBuilder().Build();

var isDaprReady = await client.CheckHealthAsync();

if (isDaprReady) 
{
    // Execute Dapr dependent code.
}
```

msfussell:
Building on the comment from @olitomlinson this proposal need clarity on what the goal is and how this works. In the update it has lost the end user viewpoint to this capability. How, why and when to use this API. 

This is my understanding:

1) A `ping` method is present on the [component API today](https://github.com/dapr/components-contrib/blob/master/health/pinger.go) and components can implement this by a component author. (This is very unfortunate name since the `ping` method is really about component health as the package describes). IMO this should `ping` method should be implemented by default on every component to reflect its initialization status (success or failure). Thereby not putting the burden on future authors and reworking existing components and also prevents the ambiguity of a component not having any health status

2) This proposal extends the Dapr side `healthz` endpoint https://docs.dapr.io/developing-applications/building-blocks/observability/sidecar-health/ to cascade health (aggregate status) across each of the components for the sidecar. Hence the code example above ^^ from Oli should just work.  
   Question: In this proposal are you expecting to call `http://localhost:3500/v1.0/healthz?include_components=true`? If so now we are going to have to update each SDK to have this additional parameter on CheckHealthAsync(); and the preferred call would then be `CheckHealthAsync(true);`or similar to include component health. Is there another way instead of updating the SDK, that you could enable component health checking?

3) It was not clear to me from this proposal how component health was measured. However I presume that a single component reporting failed health made the health of the sidecar failed (it is cascading). This is why it is important that `ping` is implemented by default IMO, for component initialization status. 

4) Given that the `/healthz` endpoint is unsecure, this proposal hints at having the metadata endpoint return additional information about failed components, but does not say that explicitly. Is that the intention? If so, what does the structure of this information look like for each healthy and unhealthy component?

5) I do not feel that we should be adding more information to the metadata endpoint without being clear on its usage. It is becoming a place to return all information, growing in size and most of this not useful for any given request. The metadata endpoint needs to be thought through on how it can be parameterized to return information say on "component health" or the "status of the specs like resiliency policies" or "actor types registered" etc and then it becomes more useful and explicit.  This can be a separate proposal.

0010-R-components-healthcheck.md
DeepanshuA: - I agree that 5 minutes seems higher. I have updated that to 30 seconds.
- I like this point but do we need another configuration for this? One more config will be too much, so rather gave logic that in case of unhealthy, it will Ping every `pingUpdateFrequency` / 3.

0010-R-components-healthcheck.md
DeepanshuA: Done

0010-R-components-healthcheck.md
```diff
@@ -15,6 +15,16 @@ There are many components in Dapr which don't yet implement Ping.
 Ping is not mandatory to be implemented by Components, which is the correct behavior, as it could lead to false positives.
 For this components health-check, components not implementing Ping will be omitted out.
 
+## Use-Case
+If a mandatory* component fails at the start-up, Dapr will terminate or will move to some non-workable state like CrashLoopBackoff etc., so `healthz` API or any other API can't be used.
+
+After Dapr has started, if any Mandatory component fails, this healthcheck can be used to determine what component has failed and accordingly some steps acam be undertaken.
```
olitomlinson: > After Dapr has started, if any Mandatory component fails, this healthcheck can be used to determine what component has failed and accordingly some steps can be undertaken.

**During a period of time where the component health check is failing, what will be the effect on the sidecars operation?**

Will it be the same effect as with `App Health Check` i.e.

[Taken from the App Healthcheck Docs](https://docs.dapr.io/developing-applications/building-blocks/observability/app-health/)

> When it detects a failure in the app’s health, Dapr stops accepting new work on behalf of the application by:

- Unsubscribing from all pub/sub subscriptions
- Stopping all input bindings
- Short-circuiting all service-invocation requests, which terminate in the Dapr runtime and are not forwarded to the application

> These changes are meant to be temporary, and Dapr resumes normal operations once it detects that the application is responsive again.

olitomlinson: **Just my initial take**

If a component health check is failed, the sidecar **should** go into the same state as with a failed `App Health Check`.

This helps to keeps a consistent model for both types of downstream health check, which then makes it easier to reason about the behaviour of the sidecar during the period of time where one or more various probes/checks are failing.

DeepanshuA: In case of App health, Dapr sidecar need to stop accepting events to process ( i.e. why subscriptions, input bindings or service invocation requests need to be stopped), as App itself is not Healthy and can't process these events. Do, Dapr sidecar doesn't know what to do with these events.

In case of a mandatory component health being reported as unhealthy, some features or All features of this component would be already un-usable. So, App can use this piece of information in a quick way to report back this status to 1. either some automated downstream to fix this issue or 2. Devops, which may consider some manual intervention.

But, here in case of a mandatory component being unhealthy, Dapr itself will not do any other operation. Here, rather App can decide to stop sending/receiving events via this component until it comes back.

olitomlinson: From my perspective, if the dapr sidecar knows that one or more mandatory components are not healthy, it makes little sense to allow the App to be invoked via PubSub/Service Invocation etc (I see this as inviting a preventable failure to occur)

Which is the same principle of App Health Checks -- if you know the App (or some downstream) is not healthy, don't invoke the App.

I absolutely see that some of those components may not always be needed, depending on the code path that is taken, so to enact the same behaviour as a failing App Health Check may seem heavy handed / aggressive.

However, I prefer an aggressive health check strategy -- "the sidecar is only as healthy as its weakest link" :)

---

The positive side effect of all of this is it will encourage operators to ensure that Components are `scoped` accordingly to the Apps that depend on them, rather than having no scope and Components being applied to every App.

0010-R-components-healthcheck.md
ItalyPaleAle: If un-healthy we should use an exponential backoff (with a limit of `pingUpdateFrequency`) IMHO, to avoid overloading a service.

msfussell:
These update help address questions. I would still like to see this described from an end user experience at the start. What would be written in the docs? Can we include this into the proposal

I agree with the observations from @olitomlinson that the sidecar needs to do more to prevent a know unhealthy status affecting the application. This proposal places to many burdens on both the component implementer and on the app IMO with too many settings.

1)Can pIng() have a default implementation that uses the component initialization for health status?
2) Why the need for mandatory verses optional component health? This is yet another thing to configure. All components should have a default ping() and then this need goes away.  Even if kept setting this via `spec.ignoreErrors `to True is obscure, and we should not overload one setting to mean something else. 
3) IMO The health of components should be set the default state on `healthz` endpoint. In other words you should exclude components with a flag not include components with a flag. I see the component health being that important to the overall health of the sidecar.
4) Re @olitomlinson's comment on whether Dapr sidecar can do more here based on the component health to not send messages etc to the app. I agree here and we should discuss this more
5) Are OK and NOT_OK the best status codes for health. Should these not indicate the health status?

JoshVanL:
Thanks for the proposal! Wanted to give a couple thoughts as well but don't want to rehash good points already raised to much-

I agree with @msfussell above that `Ping(context.Context) error` should be a mandatory function on every component interface, even if that means it is a no-op always return `nil` for some types. Status should always be included in the response if we move forward with this API. This ensures feature devs don't miss the implementation. As an aside, I think this should also be true for interfaces to implement `io.Closer` for similar reasons.

Please make it clear in the proposal that the `healthz` endpoint is an _unprivileged_ endpoint, and as such, should query the _privileged_ metadata endpoint for richer component status information.

Rather than the component `status` and `errorMessage` fields being flat to the component struct, I think they should be a child struct. This is important for keeping the API extensible and easier to parse for clients/humans. We also probably want to return the health status by default.



```go
type ComponentHealthz struct {
  Healthy bool `json:"healthy"`
  // StatusCode may or may not be appropriate for this Component to return.
  StatusCode *int `json:"status,omitempty"`
  // Message would be the human readable error message, or recovery message from unhealthy to healthy.
  Message *string `josn:"message,omitempty"`
}
```

> - To implement only http endpoint, at least for the first version of this API.

Can we not just do both? As I understand, the project is trying to move toward unifying the capabilities of both protocols anyway so this work is going to need to be done at some point.


Rather than having another config option `pingUpdateFrequency`, can each component not signal the appropriate infrequency  of polling for its particular type? This would alleviate needing both a new config option, and the set frequency being inappropriate for some component types and not others.

berndverst:
@DeepanshuA can you please put an **updated** version of the proposal into the issue description now - reconciling all feedback, so that it's easier to discuss? Reading the markdown isn't the great way to go about that.

PR #36, opened 2023-07-29
SDK resiliency timeout in ms and no per-req override.

artursouza:


mukundansundar:
+1 binding

daixiang0:
+1 binding

berndverst:
+1 binding

In Python SDK we do not currently use the env variable being changed here and we do not have any timeouts for any operations, except for HTTP service invocation for which we should maintain backwards compatibility.
For all other operations the timeout addition would be net new work which can just follow this updated proposal.

0009-S-sdk-resiliency.md
```diff
@@ -66,12 +66,11 @@ Cons:
 ### Design
 
 * `DAPR_API_MAX_RETRIES` defines the maximum number of retries, SDKs can determine which strategy will be implemented (linear, exponential backoff, etc). `0` is the default value and means no retry. `-1` or any negative value means infinite retries.
```
mikeee, line 68: Would this be improved by also adding a env var the strategy? This can obviously be unset for a default but I'd like to see the option of a customisable strategy or at least definable perhaps

berndverst: I personally think we should always use exponential backoff and call it a day. I don't think we should overcomplicate the SDK implementation and configuration options.

Maybe instead we could agree on the default exponential backoff parameters.

Those parameters can be made configurable by individual SDKs if desired, but I don't think this proposal (which was basically already implemented) should be expanded.

PR #40, opened 2023-09-19
Change `DAPR_GRPC_ENDPOINT` to infer TLS based on query parameter

JoshVanL:
In the current proposal, SDKs are meant to enable TLS on gRPC endpoints based on the scheme being `https`. This is wrong- gRPC uses the scheme to determine which [resolver](https://github.com/grpc/grpc/blob/master/doc/naming.md) to use. Not only is wrong w.r.t to the gRPC protocol, is always means users will not be able to use other resolvers (e.g. `unix://`) in conjunction with `TLS`.

This PR updates the proposal so that the `DAPR_GRPC_ENDPOINT` URL now includes an optional `?tls=<bool>` parameter. This parameter signals to the SDK whether to enable `TLS` on this endpoint or not. The SDK will continue to infer whether to use `TLS` or not based on the endpoint port when this parameter is not present. `DAPR_GRPC_ENDPOINT` defaults to port `443` if not present.

### DIFF
--- a/0008-S-sidecar-endpoint-tls.md
+++ b/0008-S-sidecar-endpoint-tls.md
@@ -67,12 +67,12 @@ Cons:
 
 ### Design
 
-* `DAPR_GRPC_ENDPOINT` defines entire endpoit for gRPC, not just host: `https://dapr-grpc.mycompany.com`
-* `DAPR_HTTP_ENDPOINT` defines entire endpoit for HTTP, not just host: `https://dapr-http.mycompany.com`
-* Port is parsed from the URL (`https://dapr.mycompany.com:8080`) or via the default port of the protocol used in the URL (80 for `http` and 443 for `https`)
+* `DAPR_GRPC_ENDPOINT` defines entire endpoint for gRPC, not just host: `dapr-grpc.mycompany.com`. No port in the URL defaults to 443.
+* `DAPR_HTTP_ENDPOINT` defines entire endpoint for HTTP, not just host: `https://dapr-http.mycompany.com`
+* Port is parsed from the hostport string (`dapr.mycompany.com:8080`) or via the default port of the protocol used in the URL (80 for `plaintext` and 443 for `TLS`)
 * `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT` can be set at the same time since some SDKs (Java, as of now) supports both protocols at the same time and app can pick which one to use.
-* `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT` must be parsed and the protocol will be used for SDK to determine if communication is over TLS (if not done automatically). In summary, `https` means secure channel.
-* Initially, only `http` and `https` protocols should be supported. Other protocols can be added in the future depending on each language support.
+* `DAPR_HTTP_ENDPOINT` must be parsed and the protocol will be used by SDK to determine if communication is over TLS (if not done automatically). In summary, `https` means secure channel.
+* `DAPR_GRPC_ENDPOINT` must be parsed and the query parameter will be used to determine whether the endpoint uses TLS. In summary, `?tls=true` means to use TLS. An empty query parameter defaults TLS based on port (`TLS` when port is 443, `plaintext` otherwise). SDKs should error on unrecognised or invalid query parameters.
 * `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT` have priority over existing `DAPR_HOST` and `DAPR_HTTP_PORT` or `DAPR_GRPC_PORT` environment variables. Application's hardcoded values passed via constructor takes priority over any environment variable. In summary, this is the priority list (highest on top):
   1. Values passed via constructor or builder method.
   2. Properties or any other language specific configuration framework.
@@ -87,6 +87,9 @@ https://github.com/dapr/java-sdk/blob/76aec01e9aa4af7a72b910d77685ddd3f0bf86f3/s
 
 * Compatability guarantees
 This feature should allow localhost definition too `http://127.0.0.1:3500`, for example.
+
+this feature should continue to allow using other resolvers other than DNS (e.g.
+`unix://`).
 
 * Deprecation / co-existence with existing functionality
 This feature takes priority over existing (inconsistent) environment variables from each SDK. If app provides a hardcoded value for Dapr endpoint (via constructor, for example), it takes priority.


ItalyPaleAle:
Isn’t this a breaking change that will break existing apps that are running unless they update the SDK?

elena-kolevska:
@ItalyPaleAle I worked on adding these env variables for Python and JS. The JS PR hasn't been merged yet, it's currently in review. The Python one was merged, but we haven't had a new release since. So, regarding these two SDKs - now is a good time to update them.

artursouza:
In case of "https", there should be no way to connect to an HTTPS endpoint without TLS. If all we need is to enable TLS to other schemes that can operate with or without TLS, then it would not be a breaking change. It would simply allow other schemes to work with TLS when the default is to work without.

My suggestion is to make `?tls=<bool>` as optional and ignore if when trying to disable TLS on a protocol that is over SSL by definition (like HTTPS).

0008-S-sidecar-endpoint-tls.md
```diff
@@ -67,12 +67,12 @@ Cons:
...
 * `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT` can be set at the same time since some SDKs (Java, as of now) supports both protocols at the same time and app can pick which one to use.
-* `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT` must be parsed and the protocol will be used for SDK to determine if communication is over TLS (if not done automatically). In summary, `https` means secure channel.
-* Initially, only `http` and `https` protocols should be supported. Other protocols can be added in the future depending on each language support.
+* `DAPR_HTTP_ENDPOINT` must be parsed and the protocol will be used by SDK to determine if communication is over TLS (if not done automatically). In summary, `https` means secure channel.
+* `DAPR_GRPC_ENDPOINT` must be parsed and the query parameter will be used to determine whether the endpoint uses TLS. In summary, `?tls=true` means to use TLS. An empty query parameter defaults TLS based on port (`TLS` when port is 443, `plaintext` otherwise). SDKs should error on unrecognised or invalid query parameters.
```
artursouza, line 75: the query param for TLS should be used to force TLS when the protocol is ambiguous. In case of HTTPS, it is not possible to disable TLS and enabling it via the TLS query param should be redundant.

artursouza (changes requested):

JoshVanL:
> In case of "https", there should be no way to connect to an HTTPS endpoint without TLS. If all we need is to enable TLS to other schemes that can operate with or without TLS, then it would not be a breaking change. It would simply allow other schemes to work with TLS when the default is to work without.

`https` is not a valid hostname resolver which is the reason for its entire removal in this updated proposal. The proposal today is actually saying that the `https` resolver is using the `dns` resolver, but _also_ to use TLS in the transport to the peer. This overloads what the name resolver (schema) is used for https://grpc.io/docs/guides/custom-name-resolution/

 > My suggestion is to make ?tls=<bool> as optional and ignore if when trying to disable TLS on a protocol that is over SSL by definition (like HTTPS).

Similarly to above, there is no name resolver which would require TLS in the transport.

0008-S-sidecar-endpoint-tls.md
JoshVanL: I don't think there is any ambiguity-

```
myhost => port=443 tls=true resolver=dns
myhost?tls=false => port=443 tls=false resolver=dns
myhost:443 => port=443 tls=true resolver=dns
myhost:1003 => port=1003 tls=false resolver=dns
myhost:1003?tls=true => port=1003 tls=true resolver=dns
unix://my.sock:1003 => port=<no concept of port> tls=false resolver=unix
unix://my.sock:1003?tls=true => port=<no concept of port> tls=true resolver=unix
```

yaron2:
+1 binding

XavierGeerinck:
+1 on this proposal. gRPC is non trivial and unclear at the moment and adding the `https` protocol makes it confusing. We should however try to leverage existing parsers in the SDKs where we can

shubham1172:
+1 binding

0008-S-sidecar-endpoint-tls.md
artursouza: OK. Can you put this in the proposal? It can serve as a reference for implementations.

JoshVanL: @artursouza Done

mukundansundar:
+1 binding

artursouza (changes requested):

0008-S-sidecar-endpoint-tls.md
```diff
@@ -88,6 +101,9 @@ https://github.com/dapr/java-sdk/blob/76aec01e9aa4af7a72b910d77685ddd3f0bf86f3/s
 * Compatability guarantees
 This feature should allow localhost definition too `http://127.0.0.1:3500`, for example.
 
+this feature should continue to allow using other resolvers other than DNS (e.g.
```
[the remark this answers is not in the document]
JoshVanL: @artursouza  No released SDK currently supports this. If this is merged now, we don’t have to do any backwards compatibility support.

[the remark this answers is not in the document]
elena-kolevska: We just released the python sdk that supports this last night.

artursouza:
I can vote on the condition we keep the backwards compatibility following the deprecation timeline of 2 releases for unsupported patterns in gRPC.

0008-S-sidecar-endpoint-tls.md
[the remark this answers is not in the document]
berndverst: @JoshVanL this would never have been merged in time AND allowed enough time for PRs to be made, reviewers to review PRs / maintainers to merge them in time for the SDK release.

We have to be reasonable in the way we treat approvers and maintainers.

berndverst:
> I can vote on the condition we keep the backwards compatibility following the deprecation timeline of 2 releases for unsupported patterns in gRPC.

+1 from me (for Python SDK) also based on this condition.

As mentioned in another comment, Python SDK is already released. Backwards compatibility should be maintained if possible.

0008-S-sidecar-endpoint-tls.md
```diff
@@ -67,18 +67,31 @@ Cons:
...
   2. Properties or any other language specific configuration framework.
   3. `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT`
   4. Existing `DAPR_HOST` (or equivalent, defaulting to `127.0.0.1`) + `DAPR_HTTP_PORT` or `DAPR_GRPC_PORT`
 
+`DAPR_GRPC_ENDPOINT` host port parsing example:
+
+```
+myhost => port=443 tls=true resolver=dns
+myhost?tls=false => port=443 tls=false resolver=dns
+myhost:443 => port=443 tls=true resolver=dns
```
artursouza, line 87: BTW, maybe we revert the decision to keep port 443 as default to TLS and keep it consistent. I am trying to think about the users that won't read the fine print and have the expected behavior.

0008-S-sidecar-endpoint-tls.md
```diff
@@ -67,18 +67,31 @@ Cons:
...
 * `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT` have priority over existing `DAPR_HOST` and `DAPR_HTTP_PORT` or `DAPR_GRPC_PORT` environment variables. Application's hardcoded values passed via constructor takes priority over any environment variable. In summary, this is the priority list (highest on top):
   1. Values passed via constructor or builder method.
   2. Properties or any other language specific configuration framework.
   3. `DAPR_GRPC_ENDPOINT` and `DAPR_HTTP_ENDPOINT`
   4. Existing `DAPR_HOST` (or equivalent, defaulting to `127.0.0.1`) + `DAPR_HTTP_PORT` or `DAPR_GRPC_PORT`
 
+`DAPR_GRPC_ENDPOINT` host port parsing example:
+
+```
+myhost => port=443 tls=true resolver=dns
```
artursouza, line 85: This one seems fair.

artursouza (changes requested):

0008-S-sidecar-endpoint-tls.md
JoshVanL: Should we also make TLS disabled by default here, since we default port 443 to no TLS?

0008-S-sidecar-endpoint-tls.md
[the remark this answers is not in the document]
JoshVanL: Done :+1:

0008-S-sidecar-endpoint-tls.md
JoshVanL: The proposal has been updated so that TLS is always disabled by default when using port 443. I think we should have the same behaviour for when the port is explicitly defined or not in the endpoint string.

elena-kolevska: I think we should default tls to true when port 443 is explicitly set.

elena-kolevska:
Here's a matrix of different possible endpoints and how they would be parsed based on this proposal. If it looks good, I can add it to the PR so it serves as a reference, and possibly as test cases for implementations.

| URL                                                          | Endpoint string to pass to grpc client    | Hostname                          | Port | Secure | Error                                                                              |
| ------------------------------------------------------------ | ----------------------------------------- | --------------------------------- | ---- | ------ | ---------------------------------------------------------------------------------- |
| :5000                                                        | dns:localhost:5000                        | localhost                         | 5000 | FALSE  |                                                                                    |
| :5000?tls=false                                              | dns:localhost:5000                        | localhost                         | 5000 | FALSE  |                                                                                    |
| :5000?tls=true                                               | dns:localhost:5000                        | localhost                         | 5000 | TRUE   |                                                                                    |
| myhost                                                       | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| myhost?tls=false                                             | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| myhost?tls=true                                              | dns:myhost:443                            | myhost                            | 443  | TRUE   |                                                                                    |
| myhost:443                                                   | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| myhost:443?tls=false                                         | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| myhost:443?tls=true                                          | dns:myhost:443                            | myhost                            | 443  | TRUE   |                                                                                    |
| [http://myhost](http://myhost)                               | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| [http://myhost?tls=false](http://myhost?tls=false)           |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=false' |
| [http://myhost?tls=true](http://myhost?tls=true)             |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=true'  |
| [http://myhost:443](http://myhost:443)                       | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| [http://myhost:443?tls=false](http://myhost:443?tls=false)   |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=false' |
| [http://myhost:443?tls=true](http://myhost:443?tls=true)     |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=true'  |
| [http://myhost:5000](http://myhost:5000)                     | dns:myhost:5000                           | myhost                            | 5000 | FALSE  |                                                                                    |
| [http://myhost:5000?tls=false](http://myhost:5000?tls=false) |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=false' |
| [http://myhost:5000?tls=true](http://myhost:5000?tls=true)   |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=true'  |
| [https://myhost:443](https://myhost:443)                     | dns:myhost:443                            | myhost                            | 443  | TRUE   |                                                                                    |
| [https://myhost:443?tls=false](https://myhost:443?tls=false) |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=false' |
| [https://myhost:443?tls=true](https://myhost:443?tls=true)   |                                           |                                   |      |        | Error: The tls query parameter is not supported for http(s) endpoints: 'tls=true'  |
| dns:myhost                                                   | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| dns:myhost?tls=false                                         | dns:myhost:443                            | myhost                            | 443  | FALSE  |                                                                                    |
| dns:myhost?tls=true                                          | dns:myhost:443                            | myhost                            | 443  | TRUE   |                                                                                    |
| dns://myauthority:53/myhost                                  | dns://myauthority:53/myhost:443           | myhost                            | 443  | FALSE  |                                                                                    |
| dns://myauthority:53/myhost?tls=false                        | dns://myauthority:53/myhost:443           | myhost                            | 443  | FALSE  |                                                                                    |
| dns://myauthority:53/myhost?tls=true                         | dns://myauthority:53/myhost:443           | myhost                            | 443  | TRUE   |                                                                                    |
| dns://myhost                                                 |                                           |                                   |      |        | Error: Invalid dns authority 'myhost' in URL 'dns://myhost'                        |
| unix://my.sock                                               | unix://my.sock                            | my.sock                           |      | FALSE  |                                                                                    |
| unix://my.sock?tls=true                                      | unix://my.sock                            | my.sock                           |      | TRUE   |                                                                                    |
| [2001:db8:1f70::999:de8:7648:6e8]                            | dns:[2001:db8:1f70::999:de8:7648:6e8]:443 | [2001:db8:1f70::999:de8:7648:6e8] | 443  | FALSE  |                                                                                    |
| dns:[2001:db8:1f70::999:de8:7648:6e8]                        | dns:[2001:db8:1f70::999:de8:7648:6e8]:443 | [2001:db8:1f70::999:de8:7648:6e8] | 443  | FALSE  |                                                                                    |
| https://[2001:db8:1f70::999:de8:7648:6e8]                    | dns:[2001:db8:1f70::999:de8:7648:6e8]:443 | [2001:db8:1f70::999:de8:7648:6e8] | 443  | TRUE   |                                                                                    |
| host:5000/v1/dapr                                            |                                           |                                   |      |        | Error: Paths are not supported for gRPC endpoints: '/v1/dapr'                      |
| host:5000/?a=1                                               |                                           |                                   |      |        | Error: Paths are not supported for gRPC endpoints: '/'                             |

artursouza (approved):

artursouza (approved):

artursouza:
@elena-kolevska Can this table be PR-ed separately? I merged to make progress on this.

PR #42, opened 2023-10-24
CIR-trust-distribution Dapr Trust Distribution proposal.

JoshVanL:
Please review and let me know what you think.


0011-CIR-trust-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+In the case of Dapr, this involves propagating PEM encoded CA bundle files to clients and servers in the cluster, which are then used to authenticate peers over TLS.
+Today, their are two methods of CA deployment to Dapr; either Sentry generated or provided by the Dapr cluster administrator.
+From the prospective of trust distribution, these two modes are functionally the same as they both result in the `dapr-trust-bundle` ConfigMap containing the CA bundle, and the `dapr-trust-bundle` Secret containing the issuer certificate chain.
+From here, trust distribution for the control plane occurs by the Operator, Placement, and Injector services reading from the mounted `dapr-trust-bundle` ConfigMap.
+Trust distribution for Daprds occurs from the Injector patching the Daprd container with an environment variable containing the CA Bundle originating from the `dapr-trust-bundle` ConfigMap.
+
+The problem with the current strategy is that once trust is distributed once (the `dapr-trust-bundle` ConfigMap and Secret is populated), the root of trust cannot change in the cluster.
```
ItalyPaleAle, line 48: There's also the problem of rotating the root CA, which is one of the goals above?

0011-CIR-trust-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+This is because trust bundles are only injected to Dapr containers at Pod creation time.
+Trust anchors are also set as environment variables whose values are static for the entire duration of a unix process, meaning they cannot be dynamically updated, for example in the event of CA root rotation.
+Today, Dapr pods will have to be restarted in order to pick-up a new trust bundle.
+It is also undefined and untested as to whether the control plane components will successfully pick-up a new trust bundle during execution; though whether they can is irrelevant as Daprds do not also support this feature.
```
ItalyPaleAle, line 52: I don't believe they do today, no

0011-CIR-trust-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Trust Distributor
+
+It is paramount that the entity that conducts the trust distribution is separate from the entity that issues identities from that root of trust, in Dapr's case this is Sentry.
+This is because trust distribution must happen out of band of identity issuance.
+An analogous to this is roots of trust of the Internet are delivered via the computers Operating System or Internet Browser, rather than fetching them from DNS servers themselves.
+Similarly for example, asset SHA hashes should be downloaded from a separate source then the assert server themselves.
+Decoupling these roles also has the benefit of improving separation of concerns between responsibilities from the identity issuer, and the trust distributor.
+
+Trust distribution will be conducted by the Operator and written to the ConfigMap `dapr-root-ca.crt` in all Namespaces.
```
ItalyPaleAle, line 64: Can this be written only to the dapr-system namespace? Eventually we will want to support multiple Dapr control planes in a cluster, and this seems to go the other direction

0011-CIR-trust-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+The Operator is a natural fit as it is not Sentry (the identity issuance server), and machinery for Kubernetes controllers already exists in the Operator today.
+ConfigMaps are a natural choice as they can be mounted by Pods & containers in Kubernetes, and trust bundles are not secrets so Secrets are not appropriate.
+There is also prior art to other projects distributing trust in this way, such as [Istio](https://github.com/istio/istio/blob/4c65649a9b116584281fadcaf8c3dd6b42d34036/istioctl/pkg/workload/workload_test.go#L340) and cert-manager's [trust-manager](https://github.com/cert-manager/trust-manager#example-bundle).
+We can also add support for writing to Kubernetes [ClusterTrustBundles](https://github.com/kubernetes/enhancements/issues/3257), though this resource is very new, and will not be available in all target Kubernetes cluster versions.
```
ItalyPaleAle, line 68: This seems to be 1.27+ (alpha), with projected volumes available in 1.28 only. I think this is wayyy too new for us to adopt any time soon.

We don't have a precise K8s support policy, but we try to go back as much as possbile and support releases that are over 1 year old.

0011-CIR-trust-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+The Operator will need to metadata watch all Namespaces and ConfigMaps in the cluster.
+The Operator should no fully inform these resources as that will massively increase the memory consumption of the Operator.
+In the event of a Namespace being created, the Operator will write the ConfigMap to that namespace.
+The Operator will also ensure that the `dapr-root-ca.crt` ConfigMap stays consistent with its local trust bundle version.
+
+### CA Rotation
+
+CA rotation can now be solved by the new trust bundle being _appended_ to the existing `dapr-root-ca.crt` Secret in the control plane namespace.
+This new bundle containing the old and new CA will be propagated to all services by the Operator, allowing for a zero downtime & graceful roll over of the CA.
+The Dapr CLI will be updated to automate this task, ensuring that the new appended trust bundle has been correctly propergated to all namespaces before writing the new CA to sentry.
```
ItalyPaleAle, line 85: Rather than the Dapr CLI, can we get this done by an API in Sentry (that is invoked for example by deploying a new YAML)? Also due to the fact that it can take a few mins.

0011-CIR-trust-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+- Dapr multi-cluster and networking federation
+
+Related issues:
+- [Multicluster Kubernetes](https://github.com/dapr/dapr/issues/3460)
+- [[Proposal] Support third-party CA - Integrate Cert Manager with Dapr](https://github.com/dapr/dapr/issues/3968)
+- [[Proposal] Automatic root certificate rotation](https://github.com/dapr/dapr/issues/5958)
+
+### Goals
+
+- Implement an active trust distribution mechanism for Dapr in Kubernetes that responds to updates.
```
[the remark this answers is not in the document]
JoshVanL: It will work outside of Kubernetes in the sense that programs will pickup new CA bundles on file, however managing the propagation of the files to the correct locations on disk is definitely a task left to the reader. The UX on Kubernetes will be a lot more automated and "one touch" as the filesystem is virtualised and consistent using ConfigMaps. For this reason, I only highlighted this mode as a goal as it wouldn't be fair to say CA rotation was first class in standalone.

0011-CIR-trust-distribution.md
```diff
@@ -0,0 +1,116 @@
...
+- Enable root rotation with no downtime.
+
+### Non-Goals
+
+- Sentry implements CA root rotation.
+- Implement external CA support.
```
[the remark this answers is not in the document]
JoshVanL: Wasn't the intention- the goal of this proposal is to _enable_ that feature to be developed in future, i.e. this proposal is a pre-req to external CA, but that this proposal isn't attempting to tacking that feature here.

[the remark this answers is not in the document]
ItalyPaleAle: Ok. Could you clarify this please?

0011-CIR-trust-distribution.md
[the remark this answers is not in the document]
ItalyPaleAle: Ok. Can you clarify this too please?

0011-CIR-trust-distribution.md
JoshVanL: I think we can tackle this with a `namespaced` mode of the trust distributor. Not writing to every namespace prevents trust being distributed at all.

0011-CIR-trust-distribution.md
JoshVanL: Makes sense :+1: I still think it's worth pointing out as something we want to support (at some point).

0011-CIR-trust-distribution.md
ItalyPaleAle: What if a new namespace is added? Will we be watching for new namespaces?

0011-CIR-trust-distribution.md
JoshVanL: Yes, I think this would be a good feature- though we need to update the CLI to perform the migration as we are supporting it already today. Or are you suggesting the CLI calls the _sentry_ API?

ItalyPaleAle: In _some way_.

I understand that having the CLI invoking a Sentry API may be a problem on its own. So that's why I mentioned perhaps the CLI can deploy a YAML that triggers the method in Sentry (via a reconciler)

0011-CIR-trust-distribution.md
JoshVanL: Yeah exactly- I have implemented similar/same mechanisms in the past:
https://github.com/cert-manager/trust-manager/blob/main/pkg/bundle/sync.go
https://github.com/cert-manager/istio-csr/blob/main/pkg/controller/configmap.go

0011-CIR-trust-distribution.md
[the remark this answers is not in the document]
JoshVanL: Done

0011-CIR-trust-distribution.md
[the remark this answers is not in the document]
JoshVanL: Done

0011-CIR-trust-distribution.md
JoshVanL: I hear you. I'm dubious about creating dedicated CRD for with the overhead that that entails or overloading an existing CRD for this purpose. Perhaps we move this to a day 2 problem?

ItalyPaleAle: What concerns me with saying "day 2 problem" is that the root cert rotation in Dapr has always been a "day 2 problem"... and that's where we are now :)

JoshVanL: True, though I suppose that at least this has a clear path forward to a concrete solution. Another point is that Sentry will not be the holder of the root certificate when using an external CA so this API is only usable in the "default" scenario, as well as not being usable in self-hosted mode.

yaron2:
+1 binding

elena-kolevska:
+1 non-binding

daixiang0:
+1 binding

0012-CIR-trust-distribution.md
```diff
@@ -0,0 +1,117 @@
+# Dapr Trust Distribution
+
+* Author(s): joshvanl
+* State: Draft
```
artursouza, line 4: Edit state to approved.

artursouza (dismissed):

artursouza:
+1 binding

yaron2 (dismissed):
Majority achieved

PR #43, opened 2023-10-30
Added proposal for finalizing distributed lock building block

ItalyPaleAle:
The distributed lock building block (dapr/dapr#3549) is currently in alpha and only partially-implemented. This is a proposal for completing the design and guiding the implementation to eventually stabilize the building block.

### DIFF
--- a/0011-B-Distributed-Lock-building-block.md
+++ b/0011-B-Distributed-Lock-building-block.md
@@ -1,0 +1,359 @@
+# Distributed Lock building block
+
+* Author(s): Alessandro Segala (ItalyPaleAle), based on @seeflood's original design
+* State: Partially implemented
+* Updated: October 2023
+
+## Overview
+
+Dapr has been including a "distributed lock" building block since version 1.8, which was originally proposed in dapr/dapr#3549. However, the current implementation is incomplete, with the original design never fully implemented, and today we are missing some key capabilities that make these APIs useful for most cases.
+
+This proposal is built on top of dapr/dapr#3549, with the goal of completing the design, rationalizing some APIs, and providing the foundation for completing the implementation.
+
+After this proposal is accepted and implemented, it should be possible to move the distributed lock building block towards a more mature (beta and eventually stable) status.
+
+## Background
+
+This proposal is meant to complete the work started in dapr/dapr#3549. Please refer to that issue for a discussion on the busines problem that the building block is trying to solve, as well as the context.
+
+Currently, as implemented the building block offers these 2 APIs:
+
+- A non-blocking `TryLock` method, which attempts to acquire a lock on a resource if it's not already locked, or fails otherwise.
+- An `Unlock` method to release a lock.
+
+### Outstanding issues
+
+As explained in [this comment](https://github.com/dapr/dapr/issues/3549#issuecomment-1484399578), the current implementation lacks some features that are required to make the building block usable for more general scenarios. In particular, we need to implement _at least_:
+
+- A blocking `Lock` method. Unlike `TryLock`, if the resource is already locked, this method blocks until it becomes available (with a maximum timeout).
+- Support for renewing locks, either explicitly (via a `RenewLock`), implicitly (as long as the app is healthy, the lock is renewed by the Dapr sidecar unless it's explicitly released), or tied to a long-lived request/stream.
+- Better response messages and status codes. For example, the APIs today return an Internal Server Error in all error cases. It's not possible, for example, to understand if `TryLock` failed because the lock was already acquired, or because of an internal error in Dapr; likewise, `Unlock` doesn't differentiate the error response based on whether the lock was invalid or an error occurred.
+- The current HTTP APIs are not RESTful, as they do not leverage the HTTP verbs to perform operations.
+
+We also have the opportunity to rationalize the concepts of lock IDs, which are called "owners" in the current APIs. Currently, the caller is responsible for providing an owner, which means the caller (the apps) need to track them.
+
+### More components
+
+Although not directly part of this proposal, which focuses on the building block's APIs, for the building block to become beta/stable we also need to add more components. Currently, the building block only support Redis (in "cluster" mode); other possible backends include:
+
+- [Kubernetes native leases](https://kubernetes.io/docs/concepts/architecture/leases/#custom-workload)
+- [etcd leases](https://etcd.io/docs/v3.5/tutorials/how-to-create-lease/)
+- Dapr Actors, which allow using any Dapr state store that can be used as actor state store (and locks are automatically cleaned up using reminders)
+
+## Implementation Details
+
+### Backwards compatibility
+
+The proposed changes are not backwards-compatible by design. Given the current APIs alpha status, Dapr makes no promises of maintaining backwards compatibility.
+
+However, since the current APIs have been present since Dapr 1.8 and haven't changed since they have been introduced, the proposal is to:
+
+- The new APIs will have version "alpha2".
+- Maintain the existing APIs as "alpha1" for one more Dapr release, showing a warning that they are deprecated and will be removed in the next Dapr version.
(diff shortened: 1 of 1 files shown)


yaron2:
Overall LGTM with a preference for option 1 + callback API for lost locks.

It would be great if we could utilize existing etags to create a generic mechanism that works across all etag supported state stores, as it would provide the following major benefits:

1. Eliminates component specific code maintenance in components-contrib
2. Eliminates maintenance of test infrastructure including conformance and certification tests for different components
3. Provides the greatest selection of components to users out of the box
4. Dramatically reduces chances of bugs by relying on a single implementation

berndverst:
@ItalyPaleAle can you also copy the proposal text directly into the description here? Makes it easier to read :)

ItalyPaleAle:
> @ItalyPaleAle can you also copy the proposal text directly into the description here? Makes it easier to read :)

I would rather not, as then they'd need to be kept in sync. I also prefer if people leave comments in the code.

0011-B-Distributed-Lock-building-block.md
```diff
@@ -0,0 +1,359 @@
...
+```proto
+service Dapr {
+  rpc Lock (LockRequest) returns (LockResponse);
+  rpc Unlock (UnlockRequest) returns (UnlockResponse);
+  rpc ForceUnlock (ForceUnlockRequest) returns (ForceUnlockResponse);
+}
+
+message LockRequest {
+  // When set to true, if the resource is already locked, the method returns right away with an error, and does not wait for the resource to become available.
+  bool immediate = 1;
```
JoshVanL, line 176: It wasn't immediately (pun not intended) clear to me what this flag did without reading the description. Could we use some variation of `wait` which might be more natural? If we want to keep the default of no wait, then perhaps `noWait`?

0011-B-Distributed-Lock-building-block.md
```diff
@@ -0,0 +1,359 @@
...
+ 	LockID  string `json:"lockID"`
+}
+
+// RenewLockRequest is a lock renewal request.
+type RenewLockRequest struct {
+ 	// ID of the resource whose lock needs to be renewed.
+ 	// This is an arbitrary value supplied by the user.
+ 	ResourceID string `json:"resourceId"`
+ 	// ID of the lock.
+ 	LockID any `json:"lockID"`
```
JoshVanL, line 116:
```suggestion
 	LockID stromg `json:"lockID"`
```

0011-B-Distributed-Lock-building-block.md
```diff
@@ -0,0 +1,359 @@
...
+type LockResponse struct {
+ 	// ID of the lock.
+ 	LockID  string `json:"lockID"`
+}
+
+// RenewLockRequest is a lock renewal request.
+type RenewLockRequest struct {
+ 	// ID of the resource whose lock needs to be renewed.
+ 	// This is an arbitrary value supplied by the user.
+ 	ResourceID string `json:"resourceId"`
```
JoshVanL, line 114: Unless I'm mistaken, isn't `LockID` all we need here?

JoshVanL, line 114: We are a bit inconsistent in the Dapr project about `Id` vs `ID` (even in this doc), and even between variable names and their json field encodings. I think we should try and be consistent moving forward.. I am firmly in the `ID` camp :slightly_smiling_face: 
FWIW, Kubernetes also goes with `ID`.

0011-B-Distributed-Lock-building-block.md
```diff
@@ -0,0 +1,359 @@
...
+ 	LockID any `json:"lockID"`
+ 	// Maximum duration for the lock (if not renewed).
+ 	// If <= 0, the lock uses the default TTL set in the component.
+ 	TTL time.Duration `json:"ttl"`
+}
+
+// UnlockRequest is a lock release request.
+type UnlockRequest struct {
+ 	// ID of the resource that is locked.
+ 	ResourceID string `json:"resourceId"`
```
JoshVanL, line 125: Again re: needed because we have the `LockID`.

0011-B-Distributed-Lock-building-block.md
```diff
@@ -0,0 +1,359 @@
...
+
+We propose two possible ways to implement the new APIs, which are both proposed below, each with pros and cons. The difference lies in how the user removes the lock.
+
+**In both cases**:
+
+- Dapr renews locks automatically in background. Apps do not need to invoke a method to renew a lock explicitly. The TTL sent in requests is only used to control how frequently locks are to be renewed (lower TTLs cause locks to be renewed more frequently, but allow for quicker detection of failed hosts).
+- If the app becomes unhealthy (when app healthchecks are enabled) or if the sidecar crashes, locks are not renewed and are left to expire when the TTL ends.
+
+> About force unlocking: this is implemented primarily to recover from situations where the lock was owned by an app that is known to have crashed. This is a potentially dangerous activity, as the previous owner of the lock is not informed (at least, not immediately) that the lock has been taken away, at least not until it tries to renew or relinquish the lock.
```
JoshVanL, line 154: I'm quite torn as to whether we should include this. I can see the use case, however the ramifications of two apps thinking they both have a lock could potentially be huge and is giving developers a big foot gun. Perhaps we hold off on exposing this API and reevaluating if it comes up as a feature request? Not being implemented by all components is also a kicker.

0011-B-Distributed-Lock-building-block.md
```diff
@@ -0,0 +1,359 @@
...
+  bool immediate = 1;
+  // Resource ID.
+  string resource_id = 2;
+  // If <= 0, uses the default TTL.
+  Duration ttl = 3;
+}
+
+message LockServerStream {
+  oneof lock_server_stream {
+    LockPing ping = 1;
```
JoshVanL, line 283: Why do we need ping? Surely the presence of the stream means that the lock is still active? Or does this mean if the network has failed downstream, worst case scenario the lock is improperly locked for is max ping interval time?

JoshVanL (changes requested):
LGTM, some comments from me.

I'm more in favor of option 2- I think from a programing perspective this feels a lot more natural as you are acquiring a lock "object" _in code_ which you can pass around etc. Are we concerned about clients not supporting HTTP/2? Are there any SDKs which don't?

0011-B-Distributed-Lock-building-block.md
ItalyPaleAle:
```suggestion
 	LockID string `json:"lockID"`
```

0011-B-Distributed-Lock-building-block.md
ItalyPaleAle: Sure

0011-B-Distributed-Lock-building-block.md
ItalyPaleAle: In the component we definitely need the lock ID. The component itself won't maintain the state - the runtime will

0011-B-Distributed-Lock-building-block.md
ItalyPaleAle: Same, the component does need the resource ID too. The state will be kept in the runtime, not component.

0011-B-Distributed-Lock-building-block.md
ItalyPaleAle: I have been on the fence myself the entire time about this.

Proposals often indicate the "desired state", and they don't have to be implemented in full at the beginning. We can always create the API in the component and then expose it later.

0011-B-Distributed-Lock-building-block.md
ItalyPaleAle: The reason for a ping is because it allows the server to send a message over the wire. Without that, if the connection drops, it's not immediately clear to the server that the connection is dropped.

PR #44, opened 2023-11-02
Distributed scheduler building block and service proposal

cicoyle:
This design proposes 2 additions:
- A Distributed Scheduler API Building Block
- A Distributed Scheduler Control Plane Service

See images for high level overview and understanding of the architecture involved.

Please review and lmk what you think 🎉


cgillum (commented):
Overall, I really like this proposal. A scheduler service that can be used to power multiple Dapr building blocks, including actor reminders, makes a ton of sense. I do have a couple questions about the proposal below, primarily from a workflow perspective.

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+
+## Current Shortfalls
+
+The **Workflows** building block is built on top of Actor Reminders, which have scale limitation issues today. The goal is to improve the performance and scale of Actor Reminder by using the distributed scheduler.
+
+Currently, Dapr users are able to use the **Publish and Subscribe** building block, but are unable to have delayed PubSub scheduling. This scheduler  service enables users to publish a message in a future specific time , for example a week from today or a specific UTC date/time.
+
+For **Service Invocation**, this building block could also benefit from a scheduler in that it would enable the scheduling of method calls between applications.
+
+As of now, Dapr does have an **input cron binding** component, which can allow users to schedule tasks. This requires the component yaml file, where users can listen on an endpoint that is scheduled. This is limited to being an input binding only. The Scheduler Service will enable the scheduling of jobs to scale across multiple replicas, while guaranteeing at most once execution of each trigger.
```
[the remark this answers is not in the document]
cicoyle: I believe the wording may be misleading here. It's meant as, a job will be executed by the Scheduler Service that 'owns' it based on the virtual partitioning, such that it won't be fired by several instances. The proposed solution enables `repeats` for jobs, to support the actor reminder logic where a job will run up to the number of `repeats` specified.

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+{
+    "dueTime": "10s",
+    "period": "R4/PT3S",
+    "ttl": "10s"
+}
+```
+
+Similar logic can be applied to a job in the following manner: 
+
+- `dueTime` => A job is expected to be scheduled as soon as it is submitted. There is no delaying a job to be scheduled, so this feature will not be supported for jobs
```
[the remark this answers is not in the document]
cicoyle: At the time, I didn't see a need for `dueTime` given a job `schedule`. If there is a need for `dueTime` to support workflows, that the job `schedule` doesn't account for, then this can be added to the Job spec 👍🏻 

I could see wanting to use `dueTime` of waiting a month before the job will fire, and then using the `schedule` as an interval. Is that a similar scenario to what you were thinking for Workflows?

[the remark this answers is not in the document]
cgillum: Yes, the addition of `dueTime` to do things like wait a month is exactly what we'd need for workflows. In the happy path for workflows, we'd never actually need to do repeated reminders. It's the ability to schedule a single reminder in the future which is critical.

0012-BIRS-distributed-scheduler.md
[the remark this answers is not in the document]
JoshVanL: I agree with @cgillum here, "at most once" and "at least once" have very specific meanings, and providing a "at most once" makes the system unsuitable for a large range of applications- workflows being a good example.

I think we should be very clear about the guarantees we are making to users with this API, and explicitly stating that in the Goals. I would be in preference of and hope we can make this system at least once.

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+##### Example Usage
+
+Users will have a job they would like scheduled. For example, an application performs a daily backup of their database. This backup task should run every day at a specific time to ensure data integrity. The user calls to schedule their job
+using the new Dapr Scheduler Building Block.
+
+Example JSON (shown below) that you can use to schedule a job by making a request to `http://localhost:<daprPort>/v1.0/job/schedule/prd-db-backup`. This request schedules a job named `prd-db-backup` to run daily for the purpose of performing a database backup. The `@daily` schedule specification indicates that the job will run once a day, specifically at midnight (for more details, refer to the Schedule table below).
+
+```Json
+{
+  "name": "prd-db-backup",
```
JoshVanL, line 74: We have the name in both the path as well as the object here. I think we should only include the name in the path to be more consistent with other APIs.

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+We will use the same exact protos from the Public Dapr API, but inside a **new**: `dapr/proto/scheduler/scheduler.proto`. However, the one difference will be in the `Job message`, where we will add:
+
+```proto
+    // Namespace for jobs to allow for multi tenancy.
+    string namespace = 5;
+
+    // The metadata associated with the job.
+    // The sidecar will create the unique `key` for storing data in the state store and pass the generated `key` along to the scheduler service for data lookup upon ‘trigger’ time later on. 
+    // The sidecar will also add metadata in order to know whether this job is registered for an actor. This is needed, as the routing mechanism for actors is different for the callback.
+    map<string,string> metadata = 6;
```
JoshVanL, line 360: Can you add the full message here? I _think_ we want this:

```proto
// Job is the definition of a job.
message Job {
  daprv1.Job job = 1;

    // Namespace of the job.
    string namespace = 2;

    // The metadata associated with the job.
    // The sidecar will create the unique `key` for storing data in the state store and pass the generated `key` along to the scheduler service for data lookup upon ‘trigger’ time later on. 
    // The sidecar will also add metadata in order to know whether this job is registered for an actor. This is needed, as the routing mechanism for actors is different for the callback.
    map<string,string> metadata = 3;
}
```

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+
+    // Allow for the daprd sidecar to register with a control plane Scheduler Service
+    rpc Register(RegisterRequest) returns (google.protobuf.Empty) {}
+}
+
+// RegisterRequest is the message to register a daprd sidecar with a control plane service. 
+message RegisterRequest {
+    // The id of the application (app_id). This is used as the key in etcd for the Scheduler Service.
+    string app_id = 1;
+
```
JoshVanL, line 380:
```suggestion
    // The id of the application (app_id). This is used as the key in etcd for the Scheduler Service.
    string app_id = 1;
    
    string namespace = 2;

```

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+    // The hostname of the daprd sidecar.
+    string hostname = 2;
+
+    // The port of the daprd sidecar.
+    int32 port = 3;
+}
+```
+
+Alternatively, there can be a streaming connection between the daprd sidecar and the Scheduler Service.
```
JoshVanL, line 389: Strongly in favor of this :+1:

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+service Scheduler {
+
+    ...<same user-facing protos defined above>...
+
+    // Allow for the daprd sidecar to register with a control plane Scheduler Service
+    rpc Register(RegisterRequest) returns (google.protobuf.Empty) {}
+}
+
+// RegisterRequest is the message to register a daprd sidecar with a control plane service. 
+message RegisterRequest {
```
JoshVanL, line 377: I am strongly against registering daprds as a one shot requests, and saving the _hostname_ of the registered app.

- There is no strong guarentee and will often be the case that a daprd will terminate without de-registering, making the record stale.
- It can be expected that daprd's hostnames change over their lifecyles in cloud native environments, also making their records broken.
- We already have an app ID & namespace name resolver in Dapr. This would duplicate that system.
- The scheduler is going to have to consistently ping registered daprds for their health causing non-trivial amounts of traffic and resource usage. Each health check to, say, 1000s daprds is going through the entire network stack every time (DNS & TCP & TLS & close).
- Managing Job request triggers -> app ID & hostnames and keeping those records accurate is a hard problem in of itself.

There is precedent in dapr today for long running streams of active daprds to control plane components with placement and operator. I think it makes sense to continue with that approach.

JoshVanL (changes requested):
Some comments from me!

cicoyle:
[linking to release 1.13](https://github.com/dapr/dapr/issues/7093)

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+    rpc TriggerJob(TriggerRequest) returns (google.protobuf.Empty) {}
+}
+
+// TriggerRequest is the message request to convey that the job is at its 'trigger' time. 
+message TriggerRequest {
+    // The job. 
+    Job job = 1; 
+
+    // The metadata associated with the job.
+    // This can contain the generated `key` for the optional state store when the daprd sidecar needs to lookup the entire data from a state store.
```
ItalyPaleAle, line 409: This proposal was [shut down a few months ago](https://github.com/dapr/dapr/issues/6424) with the reasoning that it impacts performance negatively too much. Has the thinking changed about this?

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+###### Scheduler Service APIs
+
+In a **new** `dapr/proto/scheduler/scheduler_callback.proto`, the Scheduler Service will have a 'callback' to the app in which to fulfill the job trigger execution orchestration (ie. send back the data to the app at the time of its trigger). 
+
+```proto
+// This is intended for the Scheduler Service to call back to the daprd sidecar at the time of the job schedule... ie it's trigger time.
+service SchedulerCallback {
+
+    // Callback RPC for job schedule being at 'trigger' time
+    rpc TriggerJob(TriggerRequest) returns (google.protobuf.Empty) {}
```
ItalyPaleAle, line 400: One thing you may want to consider is some sort of rate-limiting here. For example allowing a maximum number of active jobs.

Currently we don't rate-limit reminders explicitly, but given the perf bottlenecks we have in the current implementation (for example the I/O cost of executing a reminder), there's a "natural" limit there. This new design should have a much lower cost and at this stage the risk of DoS-ing an app may be real.

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+* What alternatives have been considered, and why do they not solve the problem?
+    * Placement with Actor reminders is *very* Actor specific. The goal is to have a new scheduler in Dapr that is reusable by several building blocks.
+* What advantages does this proposal have? 
+    * This design has the advantage of enabling the Scheduler Service to be implemented in different ways. 
+    * This design also enables flexibility in where the data is stored. Whether that is in etcd in full, or partially with a reference to a state store (component) of the user's choice.
+
+### Acceptance Criteria
+
+* How will success be measured? 
+    * POCs will be done to guarantee the optimal performance solution for the Scheduler Service, testing with 3 & 5 Scheduler Service instances
```
ItalyPaleAle, line 449: Normally, 5 for an etcd cluster is considered a bit of an overkill as it offers more resiliency (can tolerate the failure of 2 nodes instead of just 1) but not necessarily better performance. In fact, compared to 3 instances, [it may even be slower when writing](https://stackoverflow.com/questions/69178090/does-scaling-etcd-affect-write-performance)

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,463 @@
...
+        * Actor Reminders
+* What alternatives have been considered, and why do they not solve the problem?
+    * Placement with Actor reminders is *very* Actor specific. The goal is to have a new scheduler in Dapr that is reusable by several building blocks.
+* What advantages does this proposal have? 
+    * This design has the advantage of enabling the Scheduler Service to be implemented in different ways. 
+    * This design also enables flexibility in where the data is stored. Whether that is in etcd in full, or partially with a reference to a state store (component) of the user's choice.
+
+### Acceptance Criteria
+
+* How will success be measured?
```
ItalyPaleAle, line 448: Please also add an acceptance criteria to test what happens in disaster scenarios (when more than n/2-1 instances fail, such as 2 out of 3). Also need to test how it recovers from total cluster shutdowns - I know of more than a few users that shut down their  development K8s clusters when not in use, for example.

0012-BIRS-distributed-scheduler.md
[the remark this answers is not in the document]
olitomlinson: +1 for adding `dueTime` -  It's a slightly different use-case to that of a `schedule` but just as important IMO, particularly for Line of Business use-cases.

Also, delayed message delivery in Azure Service Bus use an explicit UTC date time for when a message should be attempted for delivery, rather than a schedule. [example](https://learn.microsoft.com/en-us/dotnet/api/azure.messaging.servicebus.servicebusmessage.scheduledenqueuetime?view=azure-dotnet)

olitomlinson:
I'm assuming the job `name` is essentially a unique ID within a given `namespace` ?

Does the `namespace` map directly to the existing dapr `namespace` concept? Or is it something else?

For example, many building blocks that support multi-tenancy can have a scope of `{namespace}.{appId}` - State Store, PubSub and Dapr Workflows follow this pattern to ensure complete isolation.

So let's assume I have a single namespace (`ns`), and the namespace contains two Apps (App `X` and App `Y`) 

Would the following command performed on App `X` colide with App `Y` ? 
: `http://localhost:<daprPort>/v1.0/job/schedule/prd-db-backup` - Can both Apps have their own isolated job called `prd-db-backup` or will both Apps be operating on the _same_ job instance?

This may sound somewhat arbitrary, but it's worth having the discussion, because as it stands, Dapr Workflow IDs are scoped to the {namespace}.{appId} - so IMO it would make sense to be consistent by default, unless there is a reason not to.

This would also have the implication that the `/job` API operations available on the sidecar can only operate on `jobs` _owned_ by that given App ID, and consequently are not allowed to operate on Jobs that may be _owned_ by other Apps. 

Therefore the proposal should try to be more specific around scope. I can see 3 potential non-mutually exclusive options to be considered :

- `{global}` scope - can be operated on by any App in the cluster.
- `{namespace}` scope - can be operated on by _any_ App contained in a specific namespace
- `{namespace}.{appId}` - can only be operated on by a specific app contained in a specific namespace.

0012-BIRS-distributed-scheduler.md
olitomlinson: More resiliency may be worth the perf impact for some users, so I think this should still be considered.

olitomlinson:
I assume this proposal is more biased towards durability and horizontal scaling, over **precision**.

I.e. we _can_ guarantee that your job will never be invoked _before_ the schedule is due, however, we can't guarantee an ceiling time on when the job is invoked after the due time is reached.

If this is true, it's important that any literature is very clear about this as people will naturally expected perfect precision if its not explicitly stated.

0012-BIRS-distributed-scheduler.md
cicoyle: Data can be saved without limits in the default embedded db (etcd) and for flexibility users can use the state store of their choosing.

0012-BIRS-distributed-scheduler.md
cicoyle: There can be multiple disaster scenarios that are hard to predict, so I will add a backup & recovery scenario to be documented to the AC.

cicoyle:
> I'm assuming the job `name` is essentially a unique ID within a given `namespace` ?
> 
> Does the `namespace` map directly to the existing dapr `namespace` concept? Or is it something else?
> 
> For example, many building blocks that support multi-tenancy can have a scope of `{namespace}.{appId}` - State Store, PubSub and Dapr Workflows follow this pattern to ensure complete isolation.
> 
> So let's assume I have a single namespace (`ns`), and the namespace contains two Apps (App `X` and App `Y`)
> 
> Would the following command performed on App `X` colide with App `Y` ? : `http://localhost:<daprPort>/v1.0/job/schedule/prd-db-backup` - Can both Apps have their own isolated job called `prd-db-backup` or will both Apps be operating on the _same_ job instance?
> 
> This may sound somewhat arbitrary, but it's worth having the discussion, because as it stands, Dapr Workflow IDs are scoped to the {namespace}.{appId} - so IMO it would make sense to be consistent by default, unless there is a reason not to.
> 
> This would also have the implication that the `/job` API operations available on the sidecar can only operate on `jobs` _owned_ by that given App ID, and consequently are not allowed to operate on Jobs that may be _owned_ by other Apps.
> 
> Therefore the proposal should try to be more specific around scope. I can see 3 potential non-mutually exclusive options to be considered :
> 
> * `{global}` scope - can be operated on by any App in the cluster.
> * `{namespace}` scope - can be operated on by _any_ App contained in a specific namespace
> * `{namespace}.{appId}` - can only be operated on by a specific app contained in a specific namespace.

All jobs will be namespaced to the app/sidecar namespace. We will not support global jobs, unless all the jobs explicitly fall under the same namespace. There is no broadcast call back to all apps in a namespace with this proposal. That might lean towards using the PubSub building block (as a future implementation detail).

cicoyle:
> I assume this proposal is more biased towards durability and horizontal scaling, over **precision**.
> 
> I.e. we _can_ guarantee that your job will never be invoked _before_ the schedule is due, however, we can't guarantee an ceiling time on when the job is invoked after the due time is reached.
> 
> If this is true, it's important that any literature is very clear about this as people will naturally expected perfect precision if its not explicitly stated.

You should see this explicitly in the proposal now - specifically in the `Goals` section. Thanks 👍🏻

0012-BIRS-distributed-scheduler.md
ItalyPaleAle: There can be multiple failure scenarios but the one that is most critical is failure of N/2+1 nodes or every node (so with 3 nodes, failure of 2 or 3). This can be tested easily by shutting down some of the nodes.

olitomlinson:
Does the implementation of this proposal depend upon this proposal https://github.com/dapr/proposals/pull/50 at all, or are they separate endeavours?

ItalyPaleAle:
> Does the implementation of this proposal depend upon this proposal #50 at all, or are they separate endeavours?

The implementation does have a dependency on #50

cicoyle:
> Does the implementation of this proposal depend upon this proposal #50 at all, or are they separate endeavours?

They are separate.

olitomlinson:
I've been re-reading this proposal and I have a few comments 

1. There doesn't appear to be any examples of how one may author a job _task_ i.e. the users code that actually gets invoked at the time the job executes? See below, taken from the proposal, a schedule is created which invokes a `db-backup` task, **but how is this task expressed?**

```yaml
{
  "schedule": "@daily",
  "data": {
    "task": "db-backup",
    "metadata": {
      "db_name": "my-prod-db",
      "backup_location": "/backup-dir"
    }
  }
}
```

2. It's my understanding that a motivation of this proposal is to be able to also perform **scheduled Service Invocation** and **scheduled PubSub,** and **scheduled State Store operations** yet I don't see any examples of how this is achieved via an SDK?

   _(side bar : how would scheduled service invocation work? who would receive the result? this is _exactly_ where you need Dapr Workflows as an orchestration, which already has a perfect method of doing scheduled and time-limited workloads)_

   _Which brings me onto my 3rd point..._

3. **Overlap with Workflows**

   I see a significant overlap in point 2, with things that are already achievable via Dapr Workflows. (albeit, not natively possible in the SDK (yet!), its just a few lines of code to achieve today)

   Going a step further... Dapr Workflows is the perfect language for defining schedules. This proposal uses the example of a daily backup routine, which is already achievable using a super simple [Monitor/Eternal Orchestration pattern](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-patterns/#monitor) - Temporal.io and many others are leading the way here, particularly from a Developer Experience perspective. 
   
   Anecdotally, lots of people use Azure Durable Functions for orchestrating backups & cloud infrastructure, so Dapr Workflows is in good company here for this kind of Ops stuff too.
   
   My 2c, I feel we should be advocating dapr adopters to learn how to express time-implicated workloads via Dapr Workflows which can grow with complexity, rather than pointing adopters to use a declarative job model which will struggle to grow beyond the basic repetition. 


4. **Prioritisation of effort**

    At the time of writing, we have the challenge of a Dapr Workflow implementation that can't GA until the Actor Reminders problem is solved (... and a heck of a lot has already been sunk into Dapr Workflows across the runtime, docs and several SDKs over the past year!)

   Given this challenge, I would like to recommend that the effort related to this proposal is concentrated on doing the absolute bare minimum implementation to solve the Dapr Workflow / Actor Reminder problem first, rather than prioritising the delivery of the Distributed Job Building Block and all the concerns that bringing a new Building Block online actually brings (SDK, Docs, Testing etc)

   I think its fair to say that the community deserve investment in existing Building Blocks that are trying to reach GA/Stable, such as Workflows, Crypto, Distributed Lock, before we bring online another Building Block into life.

   Just to be clear, I'm not against the Distributed Building Block API surface, however the overlap with what is easily achievable with Dapr Workflows leaves me wondering if its needed, _right now_...

yaron2:
> It's my understanding that a motivation of this proposal is to be able to also perform scheduled Service Invocation and scheduled PubSub, and scheduled State Store operations yet I don't see any examples of how this is achieved via an SDK?

This is out of scope for this proposal. The meaning behind the mention of other building blocks is that users can be notified on a given schedule and then perform things like pub/sub, service invocation and others from within their code. In the future we might be able to integrate scheduling into other building blocks, but for now this is completely out of scope

> I see a significant overlap in point 2, with things that are already achievable via Dapr Workflows. (albeit, not natively possible in the SDK (yet!), its just a few lines of code to achieve today)

The purpose of this API is to both satisfy the requirements of actor reminders and Dapr workflows as well as provide users with a simple CRON like API that doesn't require users to fully embrace a completely new programming model. Since cron jobs and scheduled reminders are a very common primitive, it would be useful for users to be able to schedule these easily without the additional overhead of a full fledged programming model. Temporal for example introduced a distributed scheduler primitive in addition to their orchestration patterns

> Given this challenge, I would like to recommend that the effort related to this proposal is concentrated on doing the absolute bare minimum implementation to solve the Dapr Workflow / Actor Reminder problem

Proposals do not deal with timelines, what goes into which milestone and priorities. Specifically, it's a yes/no decision whether a design can at any point be implemented in Dapr in the future. The actual priorities for milestones when it comes to which part of a proposal to implement comes down to the feature definition and triage stage that maintainers undergo at the beginning of each milestone

olitomlinson:
> The meaning behind the mention of other building blocks is that users can be notified on a given schedule and then perform things like pub/sub, service invocation and others from within their code

Fair. Although I still don't see any mention of how user-code is activated? I assume the schedule needs to be given a path (similar to how pubsub calls a specific HTTP endpoint for a subscription)

> Temporal for example introduced a distributed scheduler primitive in addition to their orchestration patterns

They did! 

And interestingly it would appear that they chose to dogfood their own workflows programming model to implement the higher level 'schedule' concept.

With this proposal, it appears we are doing the opposite?

https://temporal.io/blog/how-we-build-it-building-scheduled-workflows-in-temporal

olitomlinson:
> Fair. Although I still don't see any mention of how user-code is activated? I assume the schedule needs to be given a path (similar to how pubsub calls a specific HTTP endpoint for a subscription)

I still would like some clarity on how this all manifests

0012-BIRS-distributed-scheduler.md
cicoyle: Upon further implementation exploration, we decided it was best to move forward with streaming. The proposal has been updated to reflect current implementation details

cicoyle:
Note: I left the List jobs endpoint in the the proposal, however for 1.14 we will not implement this API due to more work required for pagination.

cicoyle:
Following are the performance numbers and associated improvements observed in the PoC [tracked in PR]( https://github.com/dapr/dapr/pull/7768):

With HA mode (3 scheduler instances), we are able to schedule 50,000 actor reminders with an average trigger qps of 4582. This shows at least a 10x improvement while keeping roughly the same qps. Invoking the Scheduler Job API directly, we observed a qps of ~35,000 for triggering.
When creating actor reminders with Dapr 1.13, we saw a qps of 50. With Scheduler, we see a qps of 4,000. This shows a drastic improvement (80x) in the scheduling of actor reminders with Scheduler. Overall, Scheduler is at most limited by the storage in etcd of registered reminders, but not on throughput of reminders

olitomlinson:
Great work cassie and Team!

This all seems very positive and I can't wait to try it out within the context of Workflows - Particularly running workflow Apps with more than just 1/2 instances, as this is a critical milestone/achievement to GA actors.

I'm actually on Vacay for the next 2 weeks, but I have my Mac with me, so if someone can publish a container image and some instructions how to integrate it, I will happily run it within my docker compose test harness that I've been using for load testing Workflows - DM me on Discord to discuss further :)

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,611 @@
...
+  optional uint32 repeats = 3;
+
+  // Optional: sets time at which or time interval before the callback is invoked for the first time.
+  optional string due_time = 4;
+
+  // Optional: Time To Live to allow for auto deletes (accounting for Actor Reminders).
+  optional string ttl = 5;
+
+  // Job data.
+  google.protobuf.Any data = 6;
```
mikeee, line 313: Just a question - noticed when working on the go-sdk implementation as well as @ItalyPaleAle highlighting it.
Is there a reason we're using a WKT here rather than byte string?

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,611 @@
...
+  optional uint32 repeats = 2;
+
+  // Optional: sets time at which or time interval before the callback is invoked for the first time.
+  optional string due_time = 3;
+
+  // Optional: Time To Live to allow for auto deletes (accounting for Actor Reminders).
+  optional string ttl = 4;
+
+  // Job data.
+  google.protobuf.Any data = 5;
```
mikeee, line 400: Same q as above about WKT usage

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,611 @@
...
+// WatchJobsRequestInitial is the initial request to start watching for jobs.
+message WatchJobsRequestInitial {
+  // app_id is the App ID of the requester.
+  string app_id = 1;
+
+  // namespace is the namespace of the requester.
+  string namespace = 2;
+
+  // actorTypes is the optional list of actor types to watch for.
+  repeated string actorTypes = 3;
```
mikeee, line 457:
```suggestion
  repeated string actor_types = 3;
```

I think this reads better following the style guide on field names

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,611 @@
...
+        * maximum number of reminders
+        * have a backup and recovery scenario of when Scheduler cluster permanently fails 
+        
+## Completion Checklist
+
+What changes or actions are required to make this proposal complete? Some examples:
+
+* Scheduler Building Block API code
+* Scheduler Service code
+* Tests added (e2e, unit)
```
mikeee, line 608: Will this be validated in the weekly/release longhauls as well?

mikeee (commented):
Amazing work! Loving the preliminary numbers.

ItalyPaleAle:
@cicoyle what are the perf numbers for executing reminders? The scheduling was always an issue in the current implementation, but more interestingly would be measuring how they are executed

artursouza:
> With HA mode (3 scheduler instances), we are able to schedule 50,000 actor reminders with an average trigger qps of 4582. This shows at least a 10x improvement while keeping roughly the same qps. Invoking the Scheduler Job API directly, we observed a qps of ~35,000 for triggering.
> When creating actor reminders with Dapr 1.13, we saw a qps of 50. With Scheduler, we see a qps of 4,000. This shows a drastic improvement (80x) in the scheduling of actor reminders with Scheduler. Overall, Scheduler is at most limited by the storage in etcd of registered reminders, but not on throughput of reminders


> @cicoyle what are the perf numbers for executing reminders? The scheduling was always an issue in the current implementation, but more interestingly would be measuring how they are executed

That is a great question. If by "executing reminders" you mean the reminders being invoked by the sidecar, then those numbers are present and referred as "triggers" in the comment from @cicoyle. If you are referring to something else for "executing reminders", please, clarify.

Thanks.

0012-BIRS-distributed-scheduler.md
artursouza: Answered in the call. It allows data to be serialized as byte[] or as "builtin" protobuf.

0012-BIRS-distributed-scheduler.md
artursouza: Same answer.

0012-BIRS-distributed-scheduler.md
artursouza: @mikeee is correct. See https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto

0012-BIRS-distributed-scheduler.md
artursouza: Adding new building blocks in alpha does not require longhaul tests. Even to stable has not been mentioned before. I suggest we modify LH tests to use scheduler (with reminders and jobs) as part of the stable criteria, not in alpha.

ItalyPaleAle:
I didn't understand those were the "triggers", thanks. In the past we discussed about the workflow perf test too, since that's heavily reliant on reminders

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,611 @@
...
+    TypeJob job = 1;
+    TypeActorReminder actor = 2;
+  }
+}
+
+// JobMetadata is the message used by the daprd sidecar to schedule a
+// job.
+message JobMetadata {
+  // app_id is the App ID of the requester.
+	string app_id = 1;
```
artursouza, line 430: nit: extra space

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,611 @@
...
+
+  // namespace is the namespace of the requester.
+  string namespace = 2;
+
+  // type is the type of the job.
+  JobMetadataType type = 3;
```
artursouza, line 436: Some better names (IMO):
JobMetadataDetails details;
JobMetadataExtension extension;
JobMetadataInfo info;

artursouza (changes requested):

0012-BIRS-distributed-scheduler.md
cicoyle: I will change it to `JobMetadataKind`

artursouza: That too Kubernetes-like.

artursouza: @yaron2 thoughts?

artursouza: Sounds better IMO.

yaron2: I support JobTargetMetadata

cicoyle:
When using the new scheduler service, we can confirm that we achieved the ability to increase throughput as the scale grows, in contrast to the existing reminder system that shows a performance degradation as scale increases. As expected, a decrease in some performance metrics is observed with the smallest parallelism and scale (details below), which then increases as parallelism and concurrency grow.

While testing parallel workflows with a max concurrent workflow count of 110 with 440 total number of workflow runs, we saw performance improvements to the tune of 275% for the rate of iterations. Furthermore, Scheduler can achieve a 267% improvement in both the data sent and received, proving that the Scheduler can handle a higher rate of data while significantly decreasing the maximum total time taken to complete the request by 62%. These numbers prove the horizontal scalability of workflows with Scheduler being used as the underlying reminder system.

While testing parallel workflows with a max concurrent count of between 60 & 90, we saw performance improvements of about 71%, where the existing reminder system drops by 44% in comparison, being unable to correct the performance drop as the scale of workflows continues to grow.

At 350 max concurrent workflows and 1400 iterations, we see a performance improvement of 50% higher than the existing reminder system. At the smaller scale of 30 max concurrent workflows and 300 total workflow count we observed a 27% decrease in the rate of iterations and 25% in data sent/received, with a 56% improvement in latency (being the maximum total time taken to complete the request).

As part of the PoC, we ran several performance tests with 100k and 10k workflows to examine other bottlenecks in the Workflow system, unrelated to the scheduling/reminder sub-systems. Several areas were identified as potential bottlenecks for Workflows as a whole and these should be an area of focus in coming iterations.

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,610 @@
...
+// Create and schedule a job
+rpc ScheduleJobAlpha1(ScheduleJobRequest) returns (google.protobuf.Empty) {}
+
+// Gets a scheduled job
+rpc GetJobAlpha1(GetJobRequest) returns (GetJobResponse) {}
+
+// Delete a job
+rpc DeleteJobAlpha1(DeleteJobRequest) returns (google.protobuf.Empty) {}
+
+// List all jobs by app
```
cicoyle, line 290: This will come post 1.14. Fine to leave as is, but also fine to rm for now

0012-BIRS-distributed-scheduler.md
```diff
@@ -0,0 +1,610 @@
...
+
+###### User-Facing APIs
+
+```proto
+service Dapr {
+…
+// Create and schedule a job
+rpc ScheduleJobAlpha1(ScheduleJobRequest) returns (google.protobuf.Empty) {}
+
+// Gets a scheduled job
```
cicoyle, line 284:
```suggestion
// Get a scheduled job
```

yaron2:
+1 binding

daixiang0:
I think we can continue to optimize it for perf issue.

+1 binding

mikeee:
+1 binding

JoshVanL:
Recusing myself from voting.

artursouza:
Recusing myself from voting.

yaron2 (approved):
Closing this vote as approved. Thanks to the author, all reviewers and maintainers for participating

PR #50, opened 2023-12-28
Proposal to Allow Multiple Reminders/Actors Implementation

DeepanshuA:
Proposal to Allow Multiple Versions of Reminder/Actors Implementation


ItalyPaleAle (approved):
LGTM thanks for porting this proposal to the proposals repo

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+(https://github.com/dapr/dapr/pull/7318)
+
+#### Step 2: Component Loading:
+
+This Control Plane Service may require some component, like a database etc. which it should be able to load, using Operator. But, components visible to a Control Plane service should not be visible to Daprd sidecar and vice-versa.
+Hence, just a logical grouping is required.
+
+(https://github.com/dapr/dapr/pull/7314)
```
mukundansundar, line 149: @DeepanshuA Can you add more details for this?

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
+# Allow Multiple Versions of Reminder/Actors Implementation
+
+* Author(s): Deepanshu Agarwal (@DeepanshuA), Alessandro Segala (@ItalyPaleAle)
+* State: Ready for Implementation
+* Updated: 2023-12-28
+
+## Overview
+
+As a user, I should be able to use Dapr core functionality of Actors/reminders i.e. leverage Actors/reminders interfaces BUT to be able to implement these interfaces in a new Control Plane service, which as a user, I should be allowed to design as per the way it solves issues for me.
```
yaron2, line 9: Why all CAPS in BUT?

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+This is a proposal to allow multiple versions of Reminders/Actors Implementation.
+
+## Background
+
+Currently, we see multiple issue with Reminders creation/deletion at scale and with placement Service.
```
cicoyle, line 14: nit:
```suggestion
Currently, we see multiple issues with Reminders creation/deletion at scale and with placement Service.
```

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Related Items
+
+This has already been discussed in different maintainers meetings, notably in [this issue](https://github.com/dapr/dapr/issues/7203).
+
+## Implementation Details
+
+### Design
+
+Reminders / Actors should have their Interfaces segregated in such a way that it shuld be possible to implement them in an independent way.
```
cicoyle, line 28:
```suggestion
Reminders / Actors should have their Interfaces segregated in such a way that it should be possible to implement them in an independent way.
```

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+```
+Idea behind this is that interfaces defined in pkg/actors/internal will be used to form a Building Block / Client which will implement all these methods defined in these interfaces.
+
+Whereas, methods defined for Services in protos will be implemented by respective new/existing Control Plane Service(s).
+
+And, the methods from these clients will call methods of the servers, wherever required.
+
+![actors_reminders](./resources/0011-R-multiple_reminders_actors_impl/Actors_Reminders_Interface.png)
+
+Depending on the implementation, it can be decided that what all RPCs need to be defined for a Service. Example, for a Placement Service which just creates and pushes Actor Types v/s Dapr Hosts table information to All sidecars, only table related ReportDaprStatusneeds to be as rpc.
```
cicoyle, line 88: Can you clarify what is meant by `v/s`, is that meant as versus? Either way I am not sure I understand your example sentence. Can you reword this sentence to be clearer?

```suggestion
Depending on the implementation, it can be decided that all RPCs need to be defined for a Service. For example, for the Placement Service which creates and pushes Actor Types v/s Dapr Hosts table information to all sidecars, only table related ReportDaprStatus needs to be an rpc.
```

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+Whereas, for a Service, where Actor information is Pulled by sidecars from Actors Service needs some methods like LookupActor or ReportActorDeactivation, which thus form a part of RPCs.
```
cicoyle, line 90: When you say for `a service`, do you mean _any_ control plane service? Or like any service a user creates for their own separate control plane service needs?
```suggestion
Whereas, for a Service, where Actor information is pulled by sidecars from the Actors Service, it only needs some methods like LookupActor or ReportActorDeactivation, which thus form a part, or subset, of the RPCs.
```

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+For some minor changes left in interfaces,  has been raised (https://github.com/dapr/dapr/pull/7281).
```
cicoyle, line 92: I think its worth elaborating on what changes the PR made if you are linking to it. It reads better if explained.
```suggestion
Standardization efforts have been made for the actor services, namely placement and reminders in [this PR](https://github.com/dapr/dapr/pull/7281).
```

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+3. XYZ used for Actors and Distributed Scheduler used for Reminders.
+
+So, how to decide what all to deploy in what case for user?
+
+### Steps:
+
+1. Dapr Configuration changes to depict what services to load for Control Plane and what not.
+
+2. Component loading capability, specifically by operator, for Control Plane. This is required, in case CP requires a component to be loaded.
```
cicoyle, line 112: What if the control plane does not require a component? Will this break anything if the control plane doesn't have a component required? Specifically, I am asking since the distributed scheduler proposal does not include a component in the design, and I want to ensure this is not a requirement and will not break things.

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+* Default is to set nothing in the CLI, which means use the built-in reminders (that use state stores)
+* If in the Helm chart the injector is configured with a reminders service, then it adds the CLI flag `--reminders-service <name>:<address>`, for example `--reminders-service scheduler:k8s-svc-name.svc.cluster.local:port` (the `<name>` part tells daprd which implementation to load)
+
+For placement:
+
+* If the user has set the `dapr.io/placement-host-address` annotation, that forces the injector to set `--placement-host-address` to the value passed, as is the case today
+* Otherwise, if there's no explicit dapr.io/placement-host-address annotation:
+    * If the configuration (via Helm) is to use placement, then sets `--placement-host-address` exactly as today (this is default value)
+    * If the configuration is to use another actors service, it sets `--actors-service <name>:<address>:<port>` following the same logic as the reminders svc)
```
cicoyle, line 136:
```suggestion
    * If the configuration is to use another actors service, it sets `--actors-service <name>:<address>:<port>` following the same logic as the reminders service)
```

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+This is backwards-compatible.
+
+Note: Exception is thrown when `--actors-service` or `--reminders-service` are set, if the injector is trying to inject a sidecar that use an oldeer version of Dapr, daprd that will crash because the flag doesn't exist. This is a "feature": if the cluster wants to use a different actors/reminders service, versions of Dapr that don't support that should crash.
```
cicoyle, line 139:
```suggestion
Note: Exception is thrown when `--actors-service` or `--reminders-service` are set, if the injector is trying to inject a sidecar that use an older version of Dapr, daprd that will crash because the flag doesn't exist. This is a "feature": if the cluster wants to use a different actors/reminders service, versions of Dapr that don't support that should crash.
```

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+
+(https://github.com/dapr/dapr/pull/7318)
+
+#### Step 2: Component Loading:
+
+This Control Plane Service may require some component, like a database etc. which it should be able to load, using Operator. But, components visible to a Control Plane service should not be visible to Daprd sidecar and vice-versa.
```
cicoyle, line 146: I assume nothing will break if a component is _not_ required?

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+Only interfaces in pkg/actors/internal have to be updated. A major work has already been done for them in release 1.12 already by segregating actors, reminders and timers interfaces (https://github.com/dapr/dapr/pull/6669).
+
+```go
+
+// PlacementService allows for interacting with the actor placement service.
+type PlacementService interface {
+	io.Closer
+
+	Start(context.Context) error
```
artursouza, line 45: This API seems to be biased towards an alternative implementation. Is there a way to show how the existing placement service will work with this interface? Is there a way to elevate the abstraction to a small set of methods:
- Start()
- WaitUntilReady()
- Lookup()
- RegisterActorType()

Or clarify what each of the other methods are needed for. If the interface is simply an union of the methods of each possible implementation, then the interface is not abstracting them properly and each vendor will need to add to this list of methods.

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Steps:
+
+1. Dapr Configuration changes to depict what services to load for Control Plane and what not.
+
+2. Component loading capability, specifically by operator, for Control Plane. This is required, in case CP requires a component to be loaded.
+
+#### Step 1: Configuration changes:
+
+A) Changes in charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
```
artursouza, line 116: I don't think a vendor's implementation should impact the OSS Chart. It is the vendor's responsibility to provide the Chart for their own services. Otherwise, each vendor will need to PR against the OSS for any specific of their custom implementation.

0011-R-multiple_reminders_actors_impl.md
```diff
@@ -0,0 +1,149 @@
...
+	io.Closer
+
+	Init(ctx context.Context) error
+	GetReminder(ctx context.Context, req *GetReminderRequest) (*Reminder, error)
+	CreateReminder(ctx context.Context, req *Reminder) error
+	DeleteReminder(ctx context.Context, req DeleteReminderRequest) error
+	DrainRebalancedReminders(actorType string, actorID string)
+	OnPlacementTablesUpdated(ctx context.Context)
+
+	SetResiliencyProvider(resiliency resiliency.Provider)
```
artursouza, line 73: Can't resiliency be offered at the layer above this interface?

artursouza (changes requested):

0011-R-multiple_reminders_actors_impl.md
DeepanshuA: Yeah, it won't break anything. That remains completely ok to not require a component by a CP Service.

0011-R-multiple_reminders_actors_impl.md
DeepanshuA: These interfaces are taken as it is directly from current dapr/dapr master.
Existing placement client works with this interface.
There are two methods which are not used right now: ReportActorDeactivation and HaltActorFn, but they are useful when actors lifecycle information needs to be updated in Placement service, if it maintains them.

I have added comments with these APIs.

0011-R-multiple_reminders_actors_impl.md
DeepanshuA: Yeah, it is not affected. I will remove it from list of updates. https://github.com/dapr/dapr/pull/7318 doesn't add any such change.

0011-R-multiple_reminders_actors_impl.md
DeepanshuA: Just noticed, this was removed already from dapr/dapr master interface. I have removed from here too.

PR #51, opened 2024-01-12
Adds 0011-CR-x509-svid-component-auth

JoshVanL:
Adds 0011-CR-x509-svid-component-auth.md which outlines enabling X.509 certificate authentication in Components using the Dapr SPIFFE identity.


jjcollinge:
probably a terrible idea but can we combine the 2? Pass the data in the context encrypted, pass the encryption key in the metadata? If someone leaks the context by mistake then it’s safe. If someone deliberately leaked the context after decrypting it then it’d be obvious in code review.

PR #52, opened 2024-03-05
Proposal: PubSub Subscription Streaming

JoshVanL:


0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,142 @@
...
+			},
+		},
+	})
+
+	stream.CloseSend()
+```
+
+## Completion Checklist
+
+- [ ] gRPC server implementation in daprd
```
ItalyPaleAle, line 135: I think we need a HTTP implementation too. Would probably be 2 APIs:

1. A method to receive events, which uses server-sent events
2. A method to confirm the completion of a message

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,142 @@
...
+  rpc SubscribeTopicEvents(stream SubscribeTopicEventsRequest) returns (stream TopicEventRequest) {}
+}
+
+// SubscribeTopicEventsRequest is a message containing the details for
+// subscribing to a topic via streaming.
+// The first message must always be the initial request. All subsequent
+// messages must be event responses.
+message SubscribeTopicEventsRequest {
+  oneof subscribe_topic_events_request_type {
+    SubscribeTopicEventsInitialRequest request = 1;
```
ItalyPaleAle, line 53:
```suggestion
    SubscribeTopicEventsSubscribeRequest request = 1;
```

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,142 @@
...
+Daprd will then setup the machinery to add this gRPC RPC stream to the set of subscribers.
+The request contains no route or path matching configuration as all events will be sent on this stream.
+Subscription gRPC streams are the highest priority when Daprd determines which publisher a message should be sent to.
+Only a single PubSub Topic pair may be subscribed at a single time with this API.
+If the first message sent to the server is not the initial request, the RPC will return an error.
+If any subsequent messages are not `SubscribeTopicEventsResponse` messages, the RPC will return an error.
+
+When a message is published to the topic, Daprd will send a `TopicEventRequest` message on the stream containing the message payload and metadata.
+After the application has processed the message, it will send to the server a `SubscribeTopicEventsResponse` containing the `id` of the message and the `status` of the message processing.
+Since multiple messages can be sent and processed in the application at the same time, the event `id` is used by the server to track the status of each individual event.
```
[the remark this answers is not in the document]
JoshVanL: I believe either from the pubsub itself, or a random UUID [here](https://github.com/dapr/components-contrib/blob/49afc557e4d7b836e8077017c503248f8fb9d6aa/pubsub/envelope.go#L77).

0013-RS-pubsub-subscription-streaming.md
artursouza: I am OK for this to be gRPC only since we have another way to subscribe for HTTP already.

yaron2: That's how we started with Configuration API I reckon, being gRPC only, but I would like to see a separate proposal for HTTP too, independent of this one.

ItalyPaleAle: When working on the crypto building block I was told it had to support HTTP too in the first iteration.

We also have another way to subscribe for gRPC too... But this way is arguably better (and would be better for HTTP too since it doesn't require opening ports in the app).

0013-RS-pubsub-subscription-streaming.md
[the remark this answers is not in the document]
ItalyPaleAle: Isn't that for cloud events only?

0013-RS-pubsub-subscription-streaming.md
yaron2: There's a distinction here in that this isn't a new building block but instead additional methods into an already existing building block where users have options to subscribe with HTTP already

ItalyPaleAle: https://github.com/dapr/proposals/blob/main/guides/api-design.md#updates-to-existing-apis

image

yaron2: This is exactly what I had in mind when I asked for a proposal for the HTTP APIs also, but these can be separate and independent still

0013-RS-pubsub-subscription-streaming.md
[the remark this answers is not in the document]
yaron2: We can generate a unique id per message the same way bulk pub/sub does today

0013-RS-pubsub-subscription-streaming.md
artursouza: I am in favor of a proposal for HTTP that does not use streaming, solving the problem of dynamic subscription for HTTP but still use the HTTP callback like we did for configuration API.

JoshVanL: I've updated the proposal with an HTTP equivalent, PTAL!

yaron2: cc @mukundansundar as per our discussion

yaron2 (changes requested):
Please change all endpoints (HTTP, gRPC) to be alpha

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,159 @@
...
+  // The metadata passing to pub components
+  //
+  // metadata property:
+  // - key : the key of the message.
+  map<string, string> metadata = 3 [json_name = "metadata"];
+
+  // dead_letter_topic is the topic to which messages that fail to be processed
+  // are sent.
+  optional string dead_letter_topic = 4 [json_name = "deadLetterTopic"];
+}
```
mukundansundar, line 79: can we add a `max_in_flight_messages` and `max_await_duration_ms` similar to bulk subscribe API https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-bulk/#subscribing-messages-in-bulk

But in this case it is the max number of in-flight messages that we allow before waiting for an acknowledgement or max await time before we wait for an acknowledgement.

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,159 @@
...
+Much like the gRPC API, the HTTP based WebSocket API will follow an initial request-response handshake, followed by a stream of messages to the client with status responses by the client, indexed by the message ID.
+The same proto types as using in the gRPC API (but in JSON blobs) will be used for the HTTP API.
+The server WebSocket implementation will be based on the [gorilla/websocket](https://github.com/gorilla/websocket) package, as this seems well used, understood and maintained.
+
+The HTTP streaming API will be available at the following endpoint.
+As the pubsub and topic information is in the request body, no request configuration is given in the URL.
+
+```
+GET: /v1.0/subscribe
+```
```
mukundansundar, line 148: can we have a request response sample for this endpoint?

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,159 @@
...
+
+Rough gRPC PoC implementation: https://github.com/dapr/dapr/commit/ed40c95d11b78ab9a36a4a8f755cf89336ae5a05
+
+The Dapr runtime gRPC server will implement the following new RPC and messages:
+
+```proto
+service Dapr {
+  // SubscribeTopicEvents subscribes to a PubSub topic and receives topic events
+  // from it.
+  rpc SubscribeTopicEvents(stream SubscribeTopicEventsRequest) returns (stream TopicEventRequest) {}
```
mukundansundar, line 47: later on, would we want to use the same RPC for bulk subscribe as well? 
If so should the TopicEventRequest stream be wrapped by another message so that we can extend to Bulk Subscribe event as well?

mukundansundar, line 47:
```suggestion
  rpc SubscribeTopicEvents(stream SubscribeTopicEventsRequestAck) returns (stream TopicEventRequest) {}
```

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,159 @@
...
+  // dead_letter_topic is the topic to which messages that fail to be processed
+  // are sent.
+  optional string dead_letter_topic = 4 [json_name = "deadLetterTopic"];
+}
+
+// SubscribeTopicEventsResponse is a message containing the result of a
+// subscription to a topic.
+message SubscribeTopicEventsResponse {
+  // id is the unique identifier for the subscription request.
+  string id = 1 [json_name = "id"];
```
mukundansundar, line 85: is this ID going to be the ID of the message? <cloud event ID>

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,159 @@
...
+	})
+
+	stream.CloseSend()
+```
+
+### HTTP (WebSockets)
+
+Along with a gRPC based streaming API, a WebSocket based HTTP equivalent API will be implemented.
+Much like the gRPC API, the HTTP based WebSocket API will follow an initial request-response handshake, followed by a stream of messages to the client with status responses by the client, indexed by the message ID.
+The same proto types as using in the gRPC API (but in JSON blobs) will be used for the HTTP API.
```
mukundansundar, line 140: can you add some more details on the messages and their actual JSON structure here in the document?

mukundansundar:
@JoshVanL could you also mention what changes are needed in SDK to support this?

0013-RS-pubsub-subscription-streaming.md
JoshVanL: Added `max_in_flight_messages`, however `max_await_duration_ms` would already be covered by existing resiliency policy.

0013-RS-pubsub-subscription-streaming.md
JoshVanL: `Ack` is not appropriate at the first message on this stream will be the initial request, not an acknowledgment.

0013-RS-pubsub-subscription-streaming.md
JoshVanL: Yes

JoshVanL:
@mukundansundar I'm not too sure what else needs to be added regarding SDK implementations- they need to implement the streaming client as described in the proposal .

0013-RS-pubsub-subscription-streaming.md
mukundansundar: I was more looking for a name which covers both messages.

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,195 @@
...
+				Id:     event.Id,
+				Status: &rtv1.TopicEventResponse{Status: rtv1.TopicEventResponse_SUCCESS},
+			},
+		},
+	})
+
+	stream.CloseSend()
+```
+
+### HTTP (WebSockets)
```
ItalyPaleAle, line 141: WebSockets make things a lot more complicated, including things like authz. We support HTTP/2 now so we should just use server-sent events (since we don't need bi-di streaming). It's much simpler to implement and use, and doesn't introduce a lot more dependencies on the server.

ItalyPaleAle (changes requested):

0013-RS-pubsub-subscription-streaming.md
yaron2: We can't use unidirectional streaming because the client needs to send back information to the server (Dapr), which is the status for each message received. From the auth perspective this wouldn't be different from how an app authenticates to Dapr today (via an access token, and unauthenticated by default) and even in browser to server scenarios a client access token is used.

ItalyPaleAle: The send-back can be done with regular HTTP requests. When using HTTP/2, they use the same TCP socket, so this wouldn't be an issue.

I would recommend avoiding adding another protocol as it does make a lot of things more complicated.

1. Go doesn't support WebSockets over HTTP/2, but only HTTP/1, so the app requires a separate socket: https://github.com/golang/go/issues/49918
2. When connecting to a WS using a web browser, you can't customize the headers, so Dapr authentication wouldn't be usable (https://devcenter.heroku.com/articles/websocket-security#authentication-authorization). I understand that connecting directly to Dapr through a web browser is not officially supported, but some customers do that. Hence why I mentioned auth is more challenging.
3. WebSockets also adds complexities when dealing with proxies (this may be a problem for some Diagrid services too presumably :) ), and local development requires tooling that is WebSockets-aware.

yaron2: The auth remains token based auth, but it doesn't have to pass through a header in this case. The initial handshake contains the token as described in the link you posted. This seems like a standard way for client side auth and is also used by the Azure Websockets (PubSub) service, as one example.

Re: web browser, it'd actually be great to make this HTTP path usable from web browsers, where http/2 doesn't have the same support as websockets. This will provide a clearer usage pattern where backend services can connect over gRPC while web services and/or any client that doesn't have gRPC or http/2 support can use websockets.

The only issue we face with websockets is the separate socket issue. If the same port cannot be shared with the existing HTTP server and the Websockets server than this is likely a non-starter.

ItalyPaleAle: > The auth remains token based auth, but it doesn't have to pass through a header in this case. The initial handshake contains the token as described in the link you posted.

Correct, but that's part of the application protocol we need to define, and it does add complexity.

> it'd actually be great to make this HTTP path usable from web browsers, where http/2 doesn't have the same support as websockets

[Support for HTTP/2](https://caniuse.com/?search=http%2F2) in browser is just as much as [WebSockets](https://caniuse.com/?search=websockets). In any case, HTTP/2 is a nice-to-have for long-lived connections but not a requirement (using HTTP/1 would be fine, but would require one separate TCP socket).

> The only issue we face with websockets is the separate socket issue. If the same port cannot be shared with the existing HTTP server and the Websockets server than this is likely a non-starter.

Separate **socket**, not port. So the long-lived connection requires its own TCP socket between the app and Dapr (1 TCP socket for the WebSockets long-lived stream, and then each request from the app will have its own). While in the case of HTTP/2 there's multiplexing being used, so there is at most 1 TCP socket being used (the long-lived stream can share the same socket as every request going to Dapr). This does improve efficiency. But it doesn't require a separate port - the same Go server can serve HTTP/1, HTTP/2, and WebSockets (but this only over HTTP/1 and not HTTP/2, per issue above)

olitomlinson:
A suggestion... rather than encouraging end-users to make **many** calls to `Dapr` to programmatically `add` and `remove` subscriptions throughout the lifetime of the `App`, please consider flipping this over to a `pull` model.

From an end-users perspective, it would be **safer** for the `App` to maintain its target state of desired subscriptions, and when ready, then the `App` signals to `Dapr` that it requests `Dapr` to _refresh_ (or `pull`) the subscriptions.

I've tried to express what I'm suggesting with a sequence diagram. I've kept the diagram to non-streaming to Keep It Simple, but I don't see why this pattern couldn't map to a streaming use-case as originally intended by this Proposal.

image

**Pros**
- **Builds on-top of existing concept already known and used by customers (Programmatic Subscriptions)**
- **Safer - End users are forced to maintain a single source of truth / desired target state of subscriptions for the `App`**
  - For ephemeral use-cases, the source of truth can be kept in memory of the `App`
  - For long-running use-cases (which must persist a process restart), the end-user can store the subscriptions in any persistant store, even Dapr State Store.
- **Guaranteed eventual reconciliation**
  - Reconciliation is just a call to `/dapr/subscribe` (or streaming equivalent)
   - Dapr always reconciles on init (Todays behavior)
   - **New** : Reconciliation can be forced by `App` calling `/v1.0/refresh-subscriptions` (or streaming equivalent) on `Dapr`
   - **New** : **Optionally** `Dapr` can periodically reconcile via `/dapr/subscribe`(or streaming equivalent) to ensure that progress is always moving forward to the target state.
 - **Debounce / Batch many changes to a PubSub component**
   - Debounce many change requests into a single operation, which is safer for PubSub components such as Kafka which do not like high-frequency change (rebalancing Consumer Groups) -  _The debounce rate could be configurable/adaptive to accomodate different PubSub component types_
  
  **Cons**
  - **Eventually consistent**
    - User needs to understand that the reconciliation will not be immediate, and may take _at-most_ up to the next periodic reconcile moment (if periodic reconcile is enabled)

yaron2:
> A suggestion... rather than encouraging end-users to make **many** calls to `Dapr` to programmatically `add` and `remove` subscriptions throughout the lifetime of the `App`, please consider flipping this over to a `pull` model.
> 
> From an end-users perspective, it would be **safer** for the `App` to maintain its target state of desired subscriptions, and when ready, then the `App` signals to `Dapr` that it requests `Dapr` to _refresh_ (or `pull`) the subscriptions.
> 
> I've tried to express what I'm suggesting with a sequence diagram. I've kept the diagram to non-streaming to Keep It Simple, but I don't see why this pattern couldn't map to a streaming use-case as originally intended by this Proposal.
> 
> image
> 
> **Pros**
> 
> * **Builds on-top of existing concept already known and used by customers (Programmatic Subscriptions)**
> * **Safer - End users are forced to maintain a single source of truth / desired target state of subscriptions for the `App`**
>   
>   * For ephemeral use-cases, the source of truth can be kept in memory of the `App`
>   * For long-running use-cases (which must persist a process restart), the end-user can store the subscriptions in any persistant store, even Dapr State Store.
> * **Guaranteed eventual reconciliation**
>   
>   * Reconciliation is just a call to `/dapr/subscribe` (or streaming equivalent)
>   * Dapr always reconciles on init (Todays behavior)
>   * **New** : Reconciliation can be forced by `App` calling `/v1.0/refresh-subscriptions` (or streaming equivalent) on `Dapr`
>   * **New** : **Optionally** `Dapr` can periodically reconcile via `/dapr/subscribe`(or streaming equivalent) to ensure that progress is always moving forward to the target state.
> * **Debounce / Batch many changes to a PubSub component**
>   
>   * Debounce many change requests into a single operation, which is safer for PubSub components such as Kafka which do not like high-frequency change (rebalancing Consumer Groups) -  _The debounce rate could be configurable/adaptive to accomodate different PubSub component types_
> 
> **Cons**
> 
> * **Eventually consistent**
>   
>   * User needs to understand that the reconciliation will not be immediate, and may take _at-most_ up to the next periodic reconcile moment (if periodic reconcile is enabled)

The scope of this proposal is to solve the highly requested feature to allow apps to subscribe dynamically without needing to listen on a port, thereby simplifying the user experience, elevating security and working around network limitations. Most of not all messaging SDKs including the ones used by Dapr encourage users to add subscriptions programmatically

olitomlinson:
@yaron2 would the sequence of what I've suggested not fit into the original framework of the proposal for streaming too?

yaron2:
> @yaron2 would the sequence of what I've suggested not fit into the original framework of the proposal for streaming too?

Not as long as it entails Dapr needing to reach the app. I think what you introduced should be coupled with introducing a pull based messaging mode to Dapr, which is highly useful in and as of itself, albeit needing a separate proposal

0013-RS-pubsub-subscription-streaming.md
yaron2: Ok, so if we don't need a separate port and there's a path forward for HTTP/2 support for WebSockets then I'm comfortable using WebSockets HTTP 1.1 to start with

ItalyPaleAle: The [issue for adding WebSockets over HTTP/2](https://github.com/golang/go/issues/49918) was opened in 2021 and looks like it's not being worked on. It seems that it is not a priority (understandably, as WebSockets is not as popular as it once was).

I think enabling a new communication protocol should be something not done lightly, as we'll be on the hook to support that for a very long time. There's nothing inherently wrong with HTTP/2, which is already supported and we have all tooling and test in place.

yaron2: The user experience that comes out of having a full-duplex bidi stream with the receive message/reply to server commands encapsulated within WebSockets is preferable to me than what the experience would be like with SSE, moreover the latter being limited to a text format. We can always change to HTTP2/SSE if we're seeing a reason to do so before moving the API to stable

olitomlinson:
> Not as long as it entails Dapr needing to reach the app.

I fully understand the proposals need to not open inbound ports on the App - I just used the non-streaming example to present the sequence of events between Dapr & App. This is why I wrote :

> I've kept the diagram to non-streaming to Keep It Simple, but I don't see why this pattern couldn't map to a streaming use-case as originally intended by this Proposal.

My assumption was that the same sequence of events could be transferred to grpc/WS? Was my assumption wrong? :)

0013-RS-pubsub-subscription-streaming.md
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// SubscribeTopicEventsResponse is a message containing the result of a
+// subscription to a topic.
+message SubscribeTopicEventsResponseAlpha1 {
+  // id is the unique identifier for the subscription request.
+  string id = 1 [json_name = "id"];
+
+  // status is the result of the subscription request.
+  TopicEventResponseAlpha1 status = 2 [json_name = "status"];
```
mikeee, line 93: `TopicEventResponse` refers to a response on a [published message](https://github.com/dapr/dapr/blob/b1959bd51621885ccf6bc4763093b91977f4be7f/dapr/proto/runtime/v1/appcallback.proto#L106) am I right in reading this as it being used as a response for the subscription response too? If so seems a bit too tightly coupled in my opinion

0013-RS-pubsub-subscription-streaming.md
daixiang0: Although WebSocket is not that popular, it is still very useful in full-duplex bidi stream case. I think it is a good start and we can Iterate it based on users' response and so on.

artursouza (approved):

artursouza:
+1 binding

daixiang0 (approved):
+1 binding

ItalyPaleAle:
I'm opposed until websockets are removed

JoshVanL:
Recusing myself from voting.

mukundansundar:
> I'm opposed until websockets are removed

Is there any specific reason why web sockets should not be supported?

0013-RS-pubsub-subscription-streaming.md
mukundansundar: I am open to looking at the maintainability aspect of Websockets and we could also potentially ping for the interest in the duplex stream from the community. 
But this proposal looks good for gRPC as a starter.

yaron2:
We have enough binding votes for this proposal.

PR #58, opened 2024-05-17
HTTP Metrics Path Matching

nelson-parente:
This proposal introduces a new opt-in API for path matching in Dapr HTTP metrics, allowing users to define specific paths to manage cardinality, improve observability, and maintain performance without relying on regexes.

### DIFF
--- a/20240517-R-http-metrics-path-normalization.md
+++ b/20240517-R-http-metrics-path-normalization.md
@@ -83,7 +83,7 @@ http:
 
 ```
 dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/{orderID}",status="200"} 4
-dapr_http_server_request_count{app_id="ping",method="GET",path="/unmatchedpath",status="200"} 1
+dapr_http_server_request_count{app_id="ping",method="GET",path="_",status="200"} 1
 ```
 
 - High Cardinality Without Path Normalization
--- a/20240517-R-http-metrics-path-normalization.md
+++ b/20240517-R-http-metrics-path-normalization.md
@@ -55,6 +55,75 @@ spec:
         - /users
 ```
 
+##### Examples
+
+Examples of how the Path Normalization API can be used to normalize metrics. The examples compare the metric `dapr_http_server_request_count` with the possible configuration combinations: low and high cardinality, with and without path normalization.
+
+- Low Cardinality Without Path Normalization
+
+
+```yaml
+http:
+  increasedCardinality: false
+  pathNormalization:
+    enabled: false
+```
+
+```
+dapr_http_server_request_count{app_id="ping",method="InvokeService/ping",status="200"} 5
+```
+- Low Cardinality With Path Normalization
+
+```yaml
+http:
+  increasedCardinality: false
+  pathNormalization:
+    enabled: true
+    ingress:
+    - /orders/{orderID}
+    egress:
+    - /orders/{orderID}
+```
+
+```
+dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/{orderID}",status="200"} 4
+dapr_http_server_request_count{app_id="ping",method="GET",path="/unmatchedpath",status="200"} 1
+```
+
+- High Cardinality Without Path Normalization
+
+```yaml
+http:
+  increasedCardinality: true
+  pathNormalization:
+    enabled: false
+```
+
+```
+dapr_http_server_request_count{app_id="ping",method="GET",path="/items/123456",status="200"} 1
+dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/1234",status="200"} 1
+dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/12345",status="200"} 1
+dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/123456",status="200"} 1
+dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/1234567",status="200"} 1
+```
+
+- High Cardinality With Path Normalization
+
+```yaml
+http:
+  increasedCardinality: true
+  pathNormalization:
+    enabled: true
+    ingress:
+    - /orders/{orderID}
+    egress:
+    - /orders/{orderID}
+```
+
+```
+dapr_http_server_request_count{app_id="ping",method="GET",path="/items/123456",status="200"} 1
+dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/{orderID}",status="200"} 4
+```
 #### Features
 
 - `pathNormalization.enabled` users can enable or disable path normalization through a straightforward boolean flag.
@@ -74,4 +143,5 @@ This Path Normalization API empowers users that rely on the metrics and observab
 
 - [ ] Implementation in daprd
 - [ ] API documentation
+- [ ] Integration, E2E tests
 
--- a/20240517-R-http-metrics-path-normalization.md
+++ b/20240517-R-http-metrics-path-normalization.md
@@ -59,7 +59,7 @@ spec:
 - `pathNormalization.enabled` users can enable or disable path normalization through a straightforward boolean flag.
 - `ingress/egress` users can specify paths for ingress and egress traffic.
 - The path matching will use the same patterns as the Go standard library (see https://pkg.go.dev/net/http#hdr-Patterns), ensuring reliable and well-supported path normalization.
-- Non-matched paths will be ignored, ensuring that cardinality doesn't grow uncontrolled.
+- Non-matched paths will added to a catch all bucket, ensuring that cardinality doesn't grow uncontrolled.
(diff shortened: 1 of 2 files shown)


20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,77 @@
...
+
+#### Features
+
+- `pathNormalization.enabled` users can enable or disable path normalization through a straightforward boolean flag.
+- `ingress/egress` users can specify paths for ingress and egress traffic.
+- The path matching will use the same patterns as the Go standard library (see https://pkg.go.dev/net/http#hdr-Patterns), ensuring reliable and well-supported path normalization.
+- Non-matched paths will added to a catch all bucket, ensuring that cardinality doesn't grow uncontrolled.
+
+
+This Path Normalization API empowers users that relly on the metrics and observability scrapped from low cardinality that will soon be the default, providing a controlled means to manage path cardinality. Those indifferent may opt for low cardinality, while legacy high cardinality mode remains available for alternative needs.
```
artursouza, line 65: Please, explain how this feature will behave when increasedCardinality is false (default in 1.14) and when increasedCardinality is true (high cardinality).

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,77 @@
...
+### Solution
+
+This proposal introduces an opt-in API for path normalization within Dapr HTTP metrics. The goal is to offer a way to normalize metrics without relying on CPU-intensive regex's.
+
+```yaml
+spec:
+  metric:
+    enabled: true
+    http:
+      increasedCardinality: true
```
mukundansundar, line 43: just confirming that `increasedCardinality` is by default `false` in 1.14 correct?

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,77 @@
...
+This Path Normalization API empowers users that rely on the metrics and observability scrapped from low cardinality that will soon be the default, providing a controlled means to manage path cardinality. Those indifferent may opt for low cardinality, while legacy high cardinality mode remains available for alternative needs.
+
+### Acceptance Criteria
+
+- The Path Normalization API is successfully integrated into the Dapr runtime, allowing users to enable/disable path normalization via configuration.
+
+## Completion Checklist
+
+- [ ] Implementation in daprd
+- [ ] API documentation
```
mukundansundar, line 76: Please add IT, E2E tests for the completion checklist.

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,77 @@
...
+    enabled: true
+    http:
+      increasedCardinality: true
+      pathNormalization:
+        enabled: true/false
+        ingress:
+        - /orders/{orderID}/items/{itemID}
+        - /users/{userID}
+        - /categories/{categoryID}/subcategories/{subCategoryID}
+        - /customers/{customerID}/orders/{orderID}
```
mukundansundar, line 50: can examples be given on what paths these would match and how that would translate to the end metrics that are being emitted?

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,77 @@
...
+        - /categories
+        - /users
+```
+
+#### Features
+
+- `pathNormalization.enabled` users can enable or disable path normalization through a straightforward boolean flag.
+- `pathNormalization.ingress/pathNormalization.egress` users can specify paths for ingress and egress path matching.
+
+The path matching will use the same patterns as the Go standard library (see https://pkg.go.dev/net/http#hdr-Patterns), ensuring reliable and well-supported path normalization.
```
mukundansundar, line 63: Will method matching also be done as part of this? Or only the path?

20240517-R-http-metrics-path-normalization.md
nelson-parente: yes, the default for 1.14 will be `false`

20240517-R-http-metrics-path-normalization.md
nelson-parente: Just path matching

20240517-R-http-metrics-path-normalization.md
nelson-parente: done!

20240517-R-http-metrics-path-normalization.md
nelson-parente: done! please take a look again at the Examples section :)

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,147 @@
...
+This proposal introduces an opt-in API for path normalization within Dapr HTTP metrics. The goal is to offer a way to normalize metrics without relying on CPU-intensive regex's.
+
+```yaml
+spec:
+  metric:
+    enabled: true
+    http:
+      increasedCardinality: true
+      pathNormalization:
+        enabled: true/false
```
JoshVanL, line 45: I think we should remove this option. I don't think it would ever be the case that a user would ever intentionally define `ingress` and `egress` path rules but didn't want them to be applied. This option will lead to frustrated users trying to use normalization but it is not being applied because they are missing the `enabled: true`.

JoshVanL (changes requested):

20240517-R-http-metrics-path-normalization.md
nelson-parente: updated 👍

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,139 @@
...
+  pathNormalization:
+    ingress:
+    - /orders/{orderID}
+    egress:
+    - /orders/{orderID}
+```
+
+```
+dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/{orderID}",status="200"} 4
+dapr_http_server_request_count{app_id="ping",method="GET",path="/unmatchedpath",status="200"} 1
```
artursouza, line 86: Can the default path be simply "_" or "" ? Something that has little risk to conflict with an actual path. Another option is to remove path completely.

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
+dapr_http_server_request_count{app_id="ping",method="InvokeService/ping",status="200"} 5
+```
+- Low Cardinality With Path Normalization
+
+```yaml
+http:
+  increasedCardinality: false
+  pathNormalization:
```
artursouza, line 77: I am not a fan of this name. Some alternatives that are better IMO:
- pathMatching
- paths
- pathGroups
- pathPatterns

I like `paths` or `pathMatching` better.

20240517-R-http-metrics-path-normalization.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Solution
+
+This proposal introduces an opt-in API for path normalization within Dapr HTTP metrics. The goal is to offer a way to normalize metrics without relying on CPU-intensive regex's.
+
+```yaml
+spec:
+  metric:
+    enabled: true
+    http:
```
artursouza, line 42: In addition, can we also add the following options to allow a way for users to have some fine grained control:
- `excludeVerb` (defaults to false)
- `excludeStatusCode` (defaults to false)
- `excludeDaprAPIs` (defaults to false)

artursouza (commented):
Made some other comments that I think is important.

20240517-R-http-metrics-path-normalization.md
nelson-parente: I agree that using "_" as the default path is a good approach. It is more explicit than using an empty string.

Removing the path completely could lead to inconsistencies, as it would result in some metrics having a path while others do not.

20240517-R-http-metrics-path-normalization.md
nelson-parente: I don't have strong feelings for the name `pathNormalization`.  What do you think about `PathCardinalityStandardization`, is it too long?

yaron2: Suggest patchMatching

20240517-R-http-metrics-path-normalization.md
nelson-parente: I agree those would be good configuration options but I'd leave them out of this proposal to keep the scope of the proposal focused on the path cardinality control.

yaron2:
+1 binding

artursouza:
+1 binding

artursouza (approved):

daixiang0:
+1 binding

daixiang0 (approved):

PR #60, opened 2024-06-18
feat: merge path matching ingress/egress paths

nelson-parente:
This proposal is an update to the original path matching proposal: https://github.com/dapr/proposals/pull/58

The motivation for this update is usability since it might not be clear to users what the difference between the egress and ingress paths is. This being said this update merges both paths.

### DIFF
--- a/20240517-R-http-metrics-path-matching.md
+++ b/20240517-R-http-metrics-path-matching.md
@@ -42,16 +42,10 @@ spec:
     http:
       increasedCardinality: true
       pathMatching:
-        ingress:
         - /orders/{orderID}/items/{itemID}
         - /users/{userID}
         - /categories/{categoryID}/subcategories/{subCategoryID}
         - /customers/{customerID}/orders/{orderID}
-        egress:
-        - /orders
-        - /orders/{orderID}/items/{itemID}
-        - /categories
-        - /users
 ```
 
 ##### Examples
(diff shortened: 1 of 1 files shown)


artursouza:
> This proposal is an update to the original path matching proposal: #58
> 
> The motivation for this update is usability since it might not be clear to users what the difference between the egress and ingress paths is. This being said this update merges both paths.

In addition, users do share config among apps, so the ingress and egress list might have a relatively big overlap. The docs examples showed that.

20240517-R-http-metrics-path-matching.md
```diff
@@ -75,15 +69,12 @@ dapr_http_server_request_count{app_id="ping",method="InvokeService/ping",status=
...
-    ingress:
-    - /orders/{orderID}
-    egress:
     - /orders/{orderID}
 ```
 
 ```
 dapr_http_server_request_count{app_id="ping",method="GET",path="/orders/{orderID}",status="200"} 4
-dapr_http_server_request_count{app_id="ping",method="GET",path="_",status="200"} 1
+dapr_http_server_request_count{app_id="ping",method="GET",path="/",status="200"} 1
```
artursouza, line 77: Does "/" mean the request was to "/" or it was a path not matched?

artursouza (changes requested):

20240517-R-http-metrics-path-matching.md
nelson-parente: in this case unmatched

artursouza (approved):

artursouza:
+1 binding

20240517-R-http-metrics-path-matching.md
nelson-parente: changed to empty string, do you agree?

yaron2:
+1 binding

20240517-R-http-metrics-path-matching.md
artursouza: unmatched path should not overlap with the metric of an actual call to "/". Thanks for changing this to be "". Empty string IMO better reflects what this actually is: path is not present.

msfussell:
+1 non-binding

daixiang0 (approved):
agree with @artursouza 

+1 binding

PR #64, opened 2024-09-16
Auto-Scale Scheduler

cicoyle:
auto scale scheduler proposal. 

This design is based off a white boarding session done w/ @JoshVanL, thanks! 🚀

### DIFF
--- a/20240912-R-scheduler-auto-scale.md
+++ b/20240912-R-scheduler-auto-scale.md
@@ -16,12 +16,11 @@ code will need to be updated such that it optionally runs embedded etcd on diffe
 
 In Dapr v1.14 we introduced the Scheduler control plane service to support the Jobs API along with scaling the Actor
 reminder system and the Workflow API (since it uses Actor reminders under the hood). The Scheduler encompasses running
-itself as a new process, an embedded etcd instance, and an internal cron scheduler. There are a few things needed to
-make the Scheduler a stable control plane service. This proposal will be detailing how to support the auto-scaling of
-Schedulers and secondly enhancing the Schedulers such that we optionally run etcd instances in the Schedulers under
-certain circumstances that make sense. When scaling beyond 5 etcd instances, there are little performance improvements, 
-so it makes sense for Schedulers to become stateless at this point. For example, we could have 20 Schedulers, but only
-5 of them would contain the embedded etcd.
+itself as a new process, an embedded etcd instance, and an internal cron scheduler. This proposal will be detailing how 
+to support the auto-scaling of Schedulers and secondly enhancing the Schedulers such that we optionally run etcd 
+instances in the Schedulers under certain circumstances that make sense. When scaling beyond 3 etcd instances, there is
+little performance improvement, so it makes sense for Schedulers to become stateless at this point. For example, we 
+could have 20 Schedulers, but only 3 of them would contain the embedded etcd.
 
 ## Related Items
 
@@ -121,7 +120,7 @@ message WatchHostsRequest {}
 
 ```
 
-3. **1/3/5 Schedulers with respective embedded etcd instances, beyond 5 instances run Schedulers stateless**. We want to
+3. **1 or 3 Schedulers with respective embedded etcd instances, beyond 3 instances run Schedulers stateless**. We want to
 optionally run embedded etcd in Scheduler for cases where it makes sense. Cases to consider:
 - running 1 Scheduler means running it with embedded etcd
 - running 2 Schedulers means running the first with embedded etcd, the second Scheduler will be stateless and connect to
@@ -130,13 +129,13 @@ the existing etcd
 - running 4 Schedulers means running 3 Schedulers with embedded etcd, and the last one will be stateless and connect to
 an existing etcd. Note, here there is no gain to round-robin-ing between the existing etcd instances due to the way 
 etcd works with raft under the hood.
-- running 5 Schedulers means running 5 Schedulers with embedded etcd.
-- running beyond 5 Schedulers means running 5 Schedulers with embedded etcd, and any Scheduler beyond 5 will be 
+- running 5 Schedulers means running 3 Schedulers with embedded etcd, the last two Scheduler instances will be stateless.
+- running beyond 5 Schedulers means running 3 Schedulers with embedded etcd, and any Scheduler beyond 3 will be 
 stateless and connect to an existing etcd instance.
 
 ![Stateless Schedulers Example](./resources/20240912-R-scheduler-auto-scale/statelessSchedulersExample.png)
 
-etcd must run with an odd number of replicas. In dapr, 5 will be our practical limit while running inside Schedulers. 
+etcd must run with an odd number of replicas. In dapr, 3 will be our practical limit while running inside Schedulers. 
 With the cron scheduler library and embedded etcd already separate processes in Scheduler, we just need to optionally 
 run the embedded etcd.
 
@@ -153,6 +152,6 @@ quorum loss as we dynamically add or remove stateful Schedulers.
 
 - [ ] Write the new control loop for the go-etcd-cron library to make the leadership table & job queue dynamic
 - [ ] Add WatchHosts to scheduler.proto & update code to use it while ensuring backwards compatibility
-- [ ] 1/3/5 Scheduler etcd problem where Schedulers might be stateless
+- [ ] 1 or 3 Scheduler etcd problem where Schedulers might be stateless
 - [ ] Tests
 - [ ] Document that some Schedulers might be stateless


20240912-R-scheduler-auto-scale.md
```diff
@@ -0,0 +1,110 @@
...
+## Background
+
+In Dapr v1.14 we introduced the Scheduler control plane service to support the Jobs API along with scaling the Actor
+reminder system and the Workflow API (since it uses Actor reminders under the hood). The Scheduler encompasses running
+itself as a new process, an embedded etcd instance, and an internal cron scheduler. There are a few things needed to
+make the Scheduler a stable control plane service. This proposal will be detailing how to support the auto-scaling of
+Schedulers and secondly enhancing the Schedulers such that we optionally run etcd instances in the Schedulers under
+certain circumstances that make sense. When scaling beyond 5 etcd instances, there are little performance improvements, 
+so it makes sense for Schedulers to become stateless at this point. For example, we could have 20 Schedulers, but only
+5 of them would contain the embedded etcd.
```
artursouza, line 24: I would say 3. Even 5 does not give much. 3 is good because it is the minimum number of replicas for HA, giving quorum vote even when one etcd instance is down. 5 means 2 etcd instances can be down. The higher availability comes at a cost of running with 5 schedulers at minimum. So, 3 is cheaper for auto-scaling scenarios.

artursouza: Another option is for user to pick which mode they want?
- non-HA: 1 etcd instance
- HA: 3 etcd instances
- "super HA": 5 etcd instances

20240912-R-scheduler-auto-scale.md
```diff
@@ -0,0 +1,110 @@
...
+### Related issues
+
+[Tracking Issue: Scheduler Stable](https://github.com/dapr/dapr/issues/7796) &&
+[Auto Scale Scheduler Issue](https://github.com/dapr/dapr/issues/7912).
+
+## Expectations and alternatives
+
+There will be 3 changes needed for this feature to be complete. 
+
+1. **New control loop** for the [go-etcd-cron library](https://github.com/diagridio/go-etcd-cron), such that the leadership table is able to be dynamic. This
```
artursouza, line 37: host address does not mean much as someone can run multiple scheduler instances on the same host. Why we need host address?

20240912-R-scheduler-auto-scale.md
```diff
@@ -0,0 +1,110 @@
...
+There will be 3 changes needed for this feature to be complete. 
+
+1. **New control loop** for the [go-etcd-cron library](https://github.com/diagridio/go-etcd-cron), such that the leadership table is able to be dynamic. This 
+means we need to watch the leadership table to observe a new Scheduler starting up or going down, stopping all Schedulers 
+job triggers for that moment until the leadership table is updated and synchronized. We will need to stop the job queue 
+and rebuild it again by reshuffling all the jobs to evenly split up the job ownership amongst Schedulers. Refactoring is
+needed here such that the queue logic can be stopped and restarted. The leadership table will be updated to 
+include the host address as the data such that the key in the table remains the id, but the value becomes the host address.
+
+- An optimization we can do after the initial work is done, is to keep all jobs in memory. Having a local cache
```
artursouza, line 44: This is fine. We can also have the cache *not* to include the job's data, then those can be cached only when the job is added to the schedule.

20240912-R-scheduler-auto-scale.md
```diff
@@ -0,0 +1,110 @@
...
+means we need to watch the leadership table to observe a new Scheduler starting up or going down, stopping all Schedulers 
+job triggers for that moment until the leadership table is updated and synchronized. We will need to stop the job queue 
+and rebuild it again by reshuffling all the jobs to evenly split up the job ownership amongst Schedulers. Refactoring is
+needed here such that the queue logic can be stopped and restarted. The leadership table will be updated to 
+include the host address as the data such that the key in the table remains the id, but the value becomes the host address.
+
+- An optimization we can do after the initial work is done, is to keep all jobs in memory. Having a local cache 
+would optimize building the queue faster, since we are using etcd this should be feasible. 
+
+2. **A new WatchHosts RPC** is added to `scheduler.proto`, allowing the dapr sidecar to connect to any available Scheduler
```
artursouza, line 47: Yes, this is good. Remember that there is a minimum number of replicas (1, 3 or 5), so the sidecar can us those first n instances as the "seed", knowing they should "always" be available, at least a majority of them.

20240912-R-scheduler-auto-scale.md
```diff
@@ -0,0 +1,110 @@
...
+
+- An optimization we can do after the initial work is done, is to keep all jobs in memory. Having a local cache 
+would optimize building the queue faster, since we are using etcd this should be feasible. 
+
+2. **A new WatchHosts RPC** is added to `scheduler.proto`, allowing the dapr sidecar to connect to any available Scheduler 
+instance via a ClusterIP service, rather than needing to manually track and connect to each instance as in the current 
+headless service model. This `WatchHosts` RPC dynamically provides the current set of active Scheduler instances to the 
+sidecar, keeping it updated as Schedulers scale up or down.
+
+- For **backward compatibility**, we will check to determine whether the `WatchHosts` RPC is available or not. If it
```
artursouza, line 52: At this point, I have a question that can make this have a different priority: is it possible to implement all this without users losing data or breaking API compatibility? If so, I would advocate to move this to 1.15 as this is a ton of work+risk for what we already have scoped for 1.15.

20240912-R-scheduler-auto-scale.md
```diff
@@ -0,0 +1,110 @@
...
+
+message WatchHostsResponse {
+  repeated HostPort hosts = 1;
+}
+
+message WatchHostsRequest {}
+
+```
+
+3. **1/3/5 Schedulers with respective embedded etcd instances, beyond 5 instances run Schedulers stateless**. We want to
```
artursouza, line 76: I don't think this is feasible. The problem here is that when you scale down, from 5 to 3 instances, for example. Suddenly, the etcd cluster "thinks" that it is down 2 instances of the 5 and does not know that the other 2 were decommissioned. Instead, I vote for the cluster to have a fixed number of instances that run with etcd: 1 or 3 and that is the minimum number of replicas the user can have in the statefulset.

20240912-R-scheduler-auto-scale.md
```diff
@@ -0,0 +1,110 @@
...
+etcd must run with an odd number of replicas. In dapr, 5 will be our practical limit while running inside Schedulers. 
+With the cron scheduler library and embedded etcd already separate processes in Scheduler, we just need to optionally 
+run the embedded etcd.
+
+Regarding [etcd's cluster reconfiguration operations guide](https://etcd.io/docs/v3.3/op-guide/runtime-configuration/#cluster-reconfiguration-operations),
+for adding members, or etcd instances, we will need to set the `initial-cluster-state` to be existing for the new 
+instances. We do not need to update the configuration of existing members manually, as once the new members join the 
+cluster, it will automatically be integrated into the consensus and the cluster's state will include the new member. 
+
+As noted in their documentation, we _**should**_ set the `strict-reconfig-check` flag on etcd such that we reject
```
artursouza, line 100: We still need to call `etcdctl member remove a8266ecf031671f3` to remove on scaling down. I suggest we get out of the business of auto-scaling etcd and run in 2 modes: 1 or 3.

artursouza (changes requested):
This is good. I have 2 concerns:
1. We should not be in the business of auto-scaling etcd, we should fix the first 1 or 3 scheduler instances to run with etcd. The problem is mostly with scaling down. Even going from 3 to 5 there is not a lot of value IMO.
2. This proposal is getting into a great deal of detail that makes me wonder: if this is not a breaking change, can we do it in 1.16?

20240912-R-scheduler-auto-scale.md
JoshVanL: Host address should include port here.

@cicoyle can we include the proto of the leadership table value we talked about? I think it was something like the following.

```proto
message Leader {
  uint32 observed_total = 1;
  string host = 2;
  uint32 port = 3;
}

JoshVanL: Ah, it's written further down 🤠

withinboredom: Using the host address and port is probably fairly brittle. For example, a controller (such as kubernetes) may move the workload to a different container on another machine. However, the 'node' will come back up with the same data but a different host + port combination.

From experience (running typesense in production on kubernetes), this results in a pretty brittle situation.

Instead, it is probably better to accept a "name" parameter and fallback to the hostname. In this way, it will "just work" with stateful sets in kubernetes (which sets the hostname as $POD-$NUM). Further it allows for running on the same host by setting 'name' to unique values.

JoshVanL: As a counter argument, I don’t think this is brittle as the stored and broadcast addresses are not used as any type of identifier, but instead as just routable addresses the client should connect to. Clients do not care which scheduler is what, just so long as it connects to all those that are available.

When a scheduler workload moves to another host address, the new address should be either updated with the new one (ip), or clients will re-resolve the host name to obtain the new ip (kube dns). When a Sched workload moves, all clients would have disconnected naturally, and begin the process of re-resolving or receive an update from the `WatchHosts` stream.

We prevent stale addresses from being persisted in the leadership table by an optimistic delete on Sched shutdown, and an Etcd key lease.

What we are trying to achieve here is the client needing only a single main Scheduler host address to connect to, and then discover the rest. This main address would ideally be a resolved service cluster ip that picks any Scheudler in the set, reresolving on disconnect. In self hosted, it would be the first scheduled in the set- which would always be the case at least.

withinboredom: I understand what you are saying, but that isn't how etcd/raft works; it is worth being explicit in the document. In any case, there is a failure condition in your elucidation, best shown by example:

1. a physical machine goes down or node gets disconnected
2. the lease will still be active AND in the leadership table
3. eventually the lease will be released, but
4. when the container is moved, it will come back, and if you previously had three etcd nodes, now you will have four.

It is more robust to design for the failure case and treat the failure case as the normal case. In other words, with hostname (or generic name), you end up with a static set in etcd. The address of the set is an implementation detail (defined through some 'well-known name', such as 
a cluster-ip, dns name, or whatever). If a node disappears from the network, it will eventually come back (or at least another one with the same name that can catch up). This allows etcd to do what it does best, while we focus on scheduling. There is no need for leases, leadership tables, etc.

The main issue with the proposed solution here is that you have two sources of truth: the nodes in the etcd set, and the nodes in the leadership table. You want to reduce that to one source of truth (etcd) by simply stating that the nodes in the etcd are defacto leaders. They can (and probably should) delegate that to other nodes as needed (potentially via partitions), but getting those delegations should be done via the leaders.

Hope that helps, and good luck!

JoshVanL: Thanks for the write-up @withinboredom!

I think there is a bit of conflation with the leadership table and the Etcd cluster itself- I don't think the proposal makes that clear atm. Schedulers all share an nth of the jobs to roughly share the work across the replicaset, based on a `mod` of the random Job ID with the replica total being equal to the replica ID. The purpose of the leadership table to ensure consistency of job ingestion by both checking no two replicas work on the same nth partition, and that there is no partition overlap of work of the Job space by checking all replicas agree on the total observed. This partition leadership is part of the design of _Scheduler_ and would be true if the Etcd cluster was remote to the replica set (something we want to support), or if we were using a different database technology all together. The leadership table also allows us in future to support other use cases like having an active & passive replica set (assuming a remote Etcd cluster or similar). Since we already have this durable partition leadership that matches the lifecycle of each replica today, the design is suggesting we extend its use to also include the hostname port so it can be used in broadcasting the latest address list of the replica set.

The logic around handling scaling of embedded or non-embedded, or recovery of an embedded Etcd node is orthogonal to the leadership key space.

withinboredom: In this case it is probably better to leave out the etcd stuff altogether and focus on the desired behavior and api this system is supposed to use. Leave the etcd stuff for a "default HA implementation" as theoretically it could be implemented by any database engine—even kubernetes configmaps and leases, probably.

cicoyle: I've pushed a commit that will hopefully clear up the meaning here :) Thanks for the input.

20240912-R-scheduler-auto-scale.md
cicoyle: For consistency sake, and if 5 doesn't offer much more benefit than 3, then Im fine with 3.

20240912-R-scheduler-auto-scale.md
cicoyle: I'd like to get this proposal approved this release (1.15) and implemented in 1.16 after Scheduler is stable since there are no breaking changes.

20240912-R-scheduler-auto-scale.md
cicoyle: Ive update the proposal to follow your suggestion of 1 or 3 etcd instances and not 5, where 3 is now our practical limit

JoshVanL:
+1 binding

yaron2:
+1 binding

PR #65, opened 2024-09-19
Resiliency Policy Error Code Retries

antontroshin:
Based on:
- Issue: https://github.com/dapr/dapr/issues/6683
- Ongoing PR: https://github.com/dapr/dapr/pull/7132

### DIFF
--- a/20240917-BR-resiliency-error-code-retries.md
+++ b/20240917-BR-resiliency-error-code-retries.md
@@ -1,0 +1,145 @@
+# Resiliency Policy Error Code Retries
+
+* Author(s): Anton Troshin (@antontroshin), Taction (@taction)
+* Updated: 2024-09-18
+
+## Overview
+
+This is a design proposal to provide additional functionality for Dapr Resiliency Policy Retries to be able to enforce policy only on specific response error codes.
+It only focuses on the `retries` (https://docs.dapr.io/operations/resiliency/policies/#retries) part of the policy.
+
+## Background
+
+In some applications, error codes may be used to indicate the business error, and retrying the operation might not be necessary or otherwise desirable.
+Customizing retry behavior will allow a more granular way to handle error codes that suit each use case.
+Currently, all errors are retried when the policy is applied.
+Some errors are not retryable, and subsequent calls will result in the same error, avoiding these retry calls will reduce the overall amount of requests, traffic, and errors.
+
+## Related Items
+
+https://github.com/dapr/dapr/issues/6683
+https://github.com/dapr/dapr/issues/6428
+https://github.com/dapr/dapr/issues/7697
+
+PR:
+https://github.com/dapr/dapr/pull/7132
+
+Docs:
+https://github.com/dapr/docs/issues/4254
+https://github.com/dapr/docs/issues/3859
+
+## Expectations and alternatives
+
+* What is in scope for this proposal?
+  - HTTP and gRPC Service Invocation, direct and proxied
+  - Bindings
+  - Pub/Sub
+
+## Implementation Details
+
+### Design
+
+Add a new object field to the `retries` policy Spec to allow the user to specify the error codes that should be retried.
+Separate fields for HTTP and gRPC. The new fields should be optional and will default to the existing behavior, which is to retry on all errors.
+
+### Example 1:
+In this example, the retry policy will retry **_only_** on HTTP 500 and HTTP error range 502-504 (inclusive) and gRPC error range 2-4 (inclusive).
+The rest of the errors will not be retried.
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Resiliency
+metadata:
+  name: myresiliency
+scopes:
+  - app1
+spec:
+  policies:
+    retries:
+      pubsubRetry:
+        policy: constant
+        duration: 5s
+        maxRetries: 10
+        matching:
+          httpStatusCodes: "500,502-504"
+          gRPCStatusCodes: "2-4"
+```
+
+### Example 2:
+In this example, the retry policy will retry **_only_** on gRPC error range 1-15 (inclusive).
+However, this policy will not apply to the HTTP errors, and they will be retried according to the default behavior, which is to retry on all errors.
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Resiliency
+metadata:
+  name: myresiliency
+scopes:
+  - app1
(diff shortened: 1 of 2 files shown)


20240917-BR-resiliency-error-code-retries.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Setting Format
+Both the `httpStatusCodes` and `gRPCStatusCodes` fields are of type string and optional and can be set to a comma-separated list of status codes and/or ranges of status codes.
+The range must be in the format `<start>-<end>` (inclusive). Having more than one dash in the range is not allowed.
+
+### Parsing the configuration
+
+The configuration values will be first parsed as comma-separated lists.
+Each entry in the list will be then parsed as a single status code or a range of status codes.
+Invalid entries will be logged and the application will fail to start.
```
artursouza, line 104: s/application/Dapr sidecar

artursouza (approved):

artursouza:
+1 binding

20240917-BR-resiliency-error-code-retries.md
```diff
@@ -0,0 +1,145 @@
...
+Docs:
+https://github.com/dapr/docs/issues/4254
+https://github.com/dapr/docs/issues/3859
+
+## Expectations and alternatives
+
+* What is in scope for this proposal?
+  - HTTP and gRPC Service Invocation, direct and proxied
+  - Bindings
+  - Pub/Sub
```
JoshVanL, line 36: How is pubsub included here- I thought this was for bare HTTP and gRPC. Pubsub clients use different clients AFAIK

20240917-BR-resiliency-error-code-retries.md
```diff
@@ -0,0 +1,145 @@
...
+spec:
+  policies:
+    retries:
+      pubsubRetry:
+        policy: constant
+        duration: 5s
+        maxRetries: 10
+        matching:
+          httpStatusCodes: "500,502-504"
+          gRPCStatusCodes: "2-4"
```
JoshVanL, line 65: Are we sure we want to use the integer status codes here, rather than the well known alias strings?

20240917-BR-resiliency-error-code-retries.md
```diff
@@ -0,0 +1,145 @@
...
+- HTTP: from 100 to 599
+- gRPC: from 1 to 16
+
+### Setting Format
+Both the `httpStatusCodes` and `gRPCStatusCodes` fields are of type string and optional and can be set to a comma-separated list of status codes and/or ranges of status codes.
+The range must be in the format `<start>-<end>` (inclusive). Having more than one dash in the range is not allowed.
+
+### Parsing the configuration
+
+The configuration values will be first parsed as comma-separated lists.
```
JoshVanL, line 102: Please add a note that we should add CEL validation to the CRD. The Regex would look something like `(\d+(-\d+)?)(,\d+(-\d+)?)*`.
Here is how we add validation https://book.kubebuilder.io/reference/markers/crd-validation

JoshVanL (changes requested):

20240917-BR-resiliency-error-code-retries.md
antontroshin: Done

20240917-BR-resiliency-error-code-retries.md
```diff
@@ -0,0 +1,149 @@
...
+- HTTP: from 100 to 599
+- gRPC: from 1 to 16
+
+### Setting Format
+Both the `httpStatusCodes` and `gRPCStatusCodes` fields are of type string and optional and can be set to a comma-separated list of status codes and/or ranges of status codes.
+The range must be in the format `<start>-<end>` (inclusive). Having more than one dash in the range is not allowed.
+
+### CRD Validation
+
+Both field values should be validated using Common Expression Language [CEL](https://kubernetes.io/docs/reference/using-api/cel/)
```
JoshVanL, line 102: Nit: please can we also add reference to the kube builder docs? https://book.kubebuilder.io/reference/markers/crd-validation

JoshVanL (changes requested):
Nit, otherwise lgtm

yaron2:
+1 binding

20240917-BR-resiliency-error-code-retries.md
```diff
@@ -0,0 +1,150 @@
...
+In addition, see Kubebuilder documentation for [CRD Validation](https://book.kubebuilder.io/reference/markers/crd-validation)
+
+### Parsing the configuration
+
+The configuration values will be first parsed as comma-separated lists.
+Each entry in the list will be then parsed as a single status code or a range of status codes.
+Invalid entries will be logged and the Dapr runtime will fail to start.
```
antontroshin, line 109: While adding this test, I found that this is not how it's happening today.
Any error in loading resiliency will be logged, but the `daprd` continues to run.
https://github.com/dapr/dapr/blob/8b7020a5b21419c564cc14bcbc4d4c1e019b3313/pkg/resiliency/resiliency.go#L294-L297
Changing this behavior will affect `daprd` startup and can produce unexpected results.
Have a few options on how to handle this:
1. leave as today, log and continue
2. intentional `Fatal` at the specific place when failing to parse the fields [here](https://github.com/dapr/dapr/blob/4c3efd73cdcdeddf1eb457016af92cf67d466966/pkg/resiliency/resiliency.go#L427)
3. introduce a special `error` type to fail `daprd` only on this case and skip on the rest
Thoughts?

JoshVanL: It’s unfortunate that today parsing a bad manifest will be completely hidden by the user until something goes wrong. Resiliency policy is certainly not the only offender on this, however it’s particularly bad for resiliency as the manifestation of non-consumed policy isn’t necessarily distinguishable from default behaviour unless you’re really looking for it.

Though not happy about it, I’m in favour of continuing the same behaviour as we have today, as in, ignore the policy if it is invalid, and address this issue separately.

Long term, we need to add proper validation at every API ingestion level (i.e. at kube API, validating webhook, and Dapr parading), as well as status reporting for all resources. We have a [components status reporting](https://github.com/dapr/proposals/pull/54) proposal already which likely has the most impact to users, but it should be expanded to all resources.

JoshVanL:
+1 binding

JoshVanL (approved):

PR #66, opened 2024-09-20
Scheduler Failure Policy

JoshVanL:


20240919-scheduler-failure-policy.md
```diff
@@ -0,0 +1,135 @@
...
+  }
+}
+
+// FailurePolicyDrop is a policy which drops the job tick when the job fails to
+// trigger.
+message FailurePolicyDrop {}
+
+// FailurePolicyRetry is a policy which retries the job at a consistent
+// delay when the job fails to trigger.
+message FailurePolicyConstant {
```
artursouza, line 87: This is redundant to the Cron schedule since you can just create a schedule with "every @1s"

JoshVanL: I like having a strongly typed interval as it means we don’t need to re-parse the value at every operation. I can foresee “@every 1s” being converted to the constant type as an optimisation.

cicoyle (dismissed):
+1 - I really like this addition, it is definitely needed. It will make the system more resilient and robust. I wonder if it might make sense to have a global scope people could set for all jobs rather than on a per job basis? Or even a per job type setting? Might be interesting and less granular than a per job basis. Just something I thought of that could be an add on to this

20240919-scheduler-failure-policy.md
```diff
@@ -0,0 +1,135 @@
...
+4. The workflow Actor Reminders will be updated so that they will be marked as only to be triggered once, with an appropriate retry failure policy when using Scheduler.
+
+
+## Design
+
+To begin with, we will support 3 failure policies:
+1. `Drop`: the job trigger will not be retried and the job will be marked as ticked.
+  This is the current behaviour and will continue to be the default behaviour for all jobs.
+2. `Constant`: the job will be retried as a constant time interval, up to a maximum number of retries (which could be infinite).
+3. `Schedule`: the job will be retried according to a [cron scheudler](https://github.com/diagridio/go-etcd-cron/blob/2a1c6747974627691165eb96a2ca0202285d71eb/proto/job.proto#L68), up to a maximum number of retries (which could be infinite).
```
mikeee, line 32:
```suggestion
3. `Schedule`: the job will be retried according to a [cron scheduler](https://github.com/diagridio/go-etcd-cron/blob/2a1c6747974627691165eb96a2ca0202285d71eb/proto/job.proto#L68), up to a maximum number of retries (which could be infinite).
```

20240919-scheduler-failure-policy.md
```diff
@@ -0,0 +1,135 @@
...
+  // last_attempted is the timestamp the job was last attempted to be triggered.
+  optional google.protobuf.Timestamp last_attempted = 5;
+}
+```
+
+Below are the proto definitions for the new `FailurePolicy` options.
+Both constant and cron policies include an optional max retries option to limit the number of retries according to the number of attempts.
+If max retries is unset, the Job will be retried indefinitely.
+The failure policy message is added as an optional field to the Job message.
+If unset, the failure policy of a Job is `Drop`.
```
mikeee, line 68: What would be the implications of setting the default failure policy to a 1s constant retry with 3 max attempts? Sure I think it's a good prod to set your own retry policies but I'd love some useful defaults.

mikeee (commented):
+1 lgtm, just a question on defaults/where a policy is unset

20240919-scheduler-failure-policy.md
JoshVanL: Yes I think this might be interesting, and am inclined to agree this should be the default. I do not see this as necessarily being different behaviour as today, since the "fire and forget" strategy would be more covered by whether the job is set to the staging queue rather than failure policy on the trigger. @cicoyle I think this also covers your point of having a global sane default.

WDYT @artursouza @cicoyle

yaron2: I support this as being the default

20240919-scheduler-failure-policy.md
```diff
@@ -0,0 +1,135 @@
...
+
+```proto
+// FailurePolicy defines the policy to apply when a job fails to trigger.
+message FailurePolicy {
+  // policy is the policy to apply when a job fails to trigger.
+  oneof policy {
+    FailurePolicyDrop drop = 1;
+    FailurePolicyConstant constant = 2;
+    FailurePolicyCron cron = 3;
```
artursouza, line 77: I suggest we remove the cron variant for now. It is too similar to Constant. Or remove Constant.

JoshVanL: SGTM :+1:

20240919-scheduler-failure-policy.md
JoshVanL: I have updated the default behavior to Constant with `1s` delay & 3 maximum retries.

mikeee: Swell, looks good to me 👌

mikeee (approved):

artursouza (approved):

artursouza:
+1 binding

yaron2:
+1 binding

cicoyle:
+1 binding

PR #67, opened 2024-09-22
Metrics for Dapr Error Codes

jake-engelberg:
# See [20241016-R-error-code-metrics.md](https://github.com/dapr/proposals/blob/f131c4e445d4b243c8aa0bdc9234c0700593b018/20240919-scheduler-failure-policy.md).


20241016-R-error-code-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+### Solution
+
+The feature is enabled via an opt-in flag within Dapr HTTP metrics (disabled by default). The goal of the opt-in functionality is to prevent the increased cost of additional metrics if they aren't being used.
+
+```yaml
+spec:
+  metric:
+    enabled: true
+    http:
+      errorCodeMetrics: true
```
JoshVanL, line 58: remove stutter

```suggestion
      errorCodes: true
```

JoshVanL (changes requested):

20241016-R-error-code-metrics.md
artursouza: Move this up a level since those error codes are not specific to HTTP.

20241016-R-error-code-metrics.md
```diff
@@ -0,0 +1,108 @@
...
+```
+From the user's perspective, the metric `dapr_error_code_count` could be split by `app-id` and `error_code` like so:
+```go
+dapr_error_code_count{app_id="service-a", error_code="ERR_ACTOR_RUNTIME_NOT_FOUND"} 4
```
artursouza, line 62: Also, to what interval are those counts emitted? I would include the timestamp and the time window too. The time window should probably be configurable too, so users might decide that they need it every 1min, 5min, 10min, 1h, depending on the app's needs.

artursouza (changes requested):

20241016-R-error-code-metrics.md
JoshVanL: Indeed, metric granularity is based on the scrape interval. Prometheus is a time series database so metrics are associated with a timestamp naturally. Users tune metric collection outside of Dapr on the Prometheus side.

20241016-R-error-code-metrics.md
jake-engelberg: @JoshVanL I decided to go with `recordErrorCodes` as I see other flags having verbs for clarity, lmk if you still suggest `errorCodes`.

20241016-R-error-code-metrics.md
```diff
@@ -0,0 +1,111 @@
...
+		...
+	...
+...
+```
+From the user's perspective, the metric `dapr_error_code_count` could be split by `app-id` and `error_code` like so:
+```go
+dapr_error_code_count{app_id="service-a", error_code="ERR_ACTOR_RUNTIME_NOT_FOUND"} 4
+dapr_error_code_count{app_id="service-a", error_code="ERR_ACTOR_INVOKE_METHOD"} 12
+dapr_error_code_count{app_id="service-b", error_code="ERR_ACTOR_RUNTIME_NOT_FOUND"} 55
+dapr_error_code_count{app_id="service-b", error_code="ERR_DIRECT_INVOKE"} 2
```
nelson-parente, line 68: In addition to tracking error codes, I believe users would benefit from having a type label (or similar) to categorize the errors by API or functional area. This would allow users to filter error metrics more efficiently without needing to hardcode specific error codes on their side. For example, users could retrieve all actor-related errors or pubsub-related errors for a particular app with ease.

An example of this would be:
```
dapr_error_code_count{app_id="service-b", error_code="ERR_ACTOR_RUNTIME_NOT_FOUND", type="ACTORS_API"} 55
```

Possible categories for the type label could include actors, pubsub, state, secrets, health, scheduler, operator, etc. This could improve the granularity of error tracking and make it easier to manage metrics across different Dapr APIs.

elena-kolevska: I agree that we would be providing a better developer experience by making it possible to group errors by API.
We do follow a schema in naming the error codes, so for example a prefix search on `ERR_ACTOR_RUNTIME*` would only return errors from the actor runtime, but this is not enforced and until now has been left to the developer as a convention.
We'd have to change the code to properly track the API. I already suggested something that would go in that direction in [this comment](https://github.com/dapr/dapr/pull/8192#pullrequestreview-2382399008).

alicejgibbons: I agree with @nelson-parente's point here, I think it would be useful to add a `type` or `group` for each error code, as this would allow easier querying and filtering for all Dapr metrics esp in backends. @jake-engelberg  can you add this to the proposal?

20241016-R-error-code-metrics.md
```diff
@@ -0,0 +1,111 @@
...
+		respondWithError(w, msg)
+		log.Debug(msg)
+		ErrorCodeMetrics.RecordError(msg.tag)
+		return
+	}
+...
+}
+```
+
+With `RecordError()` simply recording OpenCensus metrics like other metric recorders found in `pkg/diagnostics`:
```
elena-kolevska, line 90: Just a note that we're moving to Open Telemetry, so the specific code in the example here will change.

20241016-R-error-code-metrics.md
jake-engelberg: I have updated the proposal to account for aligning the error names to allow filtering by prefix, as well as general improvements to the implementation explanation

JoshVanL:
+1 binding

JoshVanL (approved):

PR #70, opened 2024-11-22
Proposal for actor metadata.

artursouza:


20241121-S-actor-sdk-metadata.md
```diff
@@ -0,0 +1,99 @@
...
+- Allows actor clients to include HTTP headers or gRPC metadata while keeping the protocol agnostic nature of the Actor SDK runtime.
+
+Cons:
+- Does not support metadata values per request.
+
+## Implementation Details
+
+### Design
+
+When instantiating an Actor client or Actor proxy on any of the SDKs, the user might optionally pass a map of string of strings with headers or metadata to be passed in every actor invocation coming from that actor client instance.
```
artursouza, line 63: I realized that this also needs runtime changes to propagate the HTTP headers or gRPC metadata.

PR #72, opened 2025-02-17
Actors: bi-directional streaming

JoshVanL:


20250115-actors-bi-di-stream.md
```diff
@@ -0,0 +1,182 @@
...
+    RequestGetState get_state = 6;
+    RequestExecuteStateTransaction execute_state_transaction = 7;
+    RequestPublishMessage publish_message = 8; // Actor pubsub not yet implemented.
+  }
+}
+```
+
+Each outbound `RequestRequest` message includes a `uid` field, which is a simple incrementing `uint64`.
+This field must start at `1` and increment by `1` for each new request.
+If a message is received from Daprd with a `uid` that is not `+1` of the previous message, the stream will be terminated.
```
elena-kolevska, line 109: Is this `from Daprd` or from the client?

elena-kolevska: Ah ok, that makes sense then. Thanks for clarifying 🙏

20250115-actors-bi-di-stream.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+### Actor API Requests
+
+After the initial message, the client sends API requests to Daprd or processes received message events.
+
+```proto
+message Request {
+  oneof actor_request_type {
```
WhitWaldo, line 81: There should be an error channel exposed here somewhere so that if the client requests the creation of an actor and some prerequisite on the runtime isn't met (e.g. cannot connect to the specified actor state store), such a message can be sent back to the SDK so an exception can be raised.

WhitWaldo (commented):
Need some way for the runtime to tell the SDK about connection errors at initialization so the client doesn't proceed with setup assuming everything is fine and working.

20250115-actors-bi-di-stream.md
JoshVanL: Sending the error response on initialization should be covered in the [`ResponseInitial` message](https://github.com/dapr/proposals/pull/72/files#diff-91f410445466f13b1f41aa800ad691ac699a1a487482db8366548b14a7b72964R137) sent back to the client.

JoshVanL: This will be covered by the [`ResponeResponse` message](https://github.com/dapr/proposals/pull/72/files#diff-91f410445466f13b1f41aa800ad691ac699a1a487482db8366548b14a7b72964R143). I'm not sure whether the error should be a generic error which lives outside the `oneof`s, or needs to be specific to each `oneof` API type. I'm assuming it needs to be inside each `oneof` message so that it can be fully typed.

WhitWaldo: Especially as this channel is potentially open for a little while, it'd be better to have a `oneof` in place so the SDK can understand if the error is transient and something that just needs the channel re-opened for or if it's more of a fatal error (and should just be logged and given up on).

I'll have to find the issue, but it's a complaint on one of the recent building blocks that when there's a failure, the SDK doesn't handle it really well because there's no real information from the runtime what to do about it except that something went wrong. It'd be great to correct that scenario here.

PR #73, opened 2025-02-17
Adds 20250117-actors-pubsub.md

JoshVanL:


20250117-actors-pubsub.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+Publishing follows the same topic naming convention:
+
+```
+dapr.actors||$namespace||$actortype||$topic-name
+```
+
+Messages published by an actor will be delivered to all actors of the same type that are subscribed to the given topic.
```
yaron2, line 91: That's a great broadcast feature. It's clear how publishing messages from actors to other actors works, can you elaborate on what it would look like for a non-actor publisher to publish a message to an actor?

JoshVanL: Thanks @yaron2, forgot that one out!

I've updated the proposal with a _new_ RPC for non-actor clients. Added a note that this RPC will not instantiate that actor to exist, since it needs to fist subscribe.

JoshVanL: Also added a comment around deactivation, and actor idle timeout.

olitomlinson:
> This feature should be implemented after the actor bi-directional streaming feature is completed.
The bi-directional stream will serve as the transport mechanism for delivering messages to actors.

Is there a proposal for this feature? As it could be considered alongside the desire for true async message passing (inbox / outbox) between Actors, as I outlined here https://github.com/dapr/dapr/issues/501#issuecomment-1689787965

20250117-actors-pubsub.md
```diff
@@ -0,0 +1,140 @@
...
+metadata:
+  - name: actorpubsub
+    value: "true"
+```
+
+Only one PubSub component can be marked as the actor PubSub.
+
+### Actor Subscriptions
+
+Actors will be able to subscribe to messages within their type for a particular topic.
```
WhitWaldo, line 35: I'd like to see this supplemented with support for [content-based routing](https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-route-messages/). While a general broadcast is certainly a much-needed feature, I can easily imagine a chatty topic activating actors unnecessarily. Limiting such activations to only those actors registered to receive that subset would be a solid (and likely expected) part of this functionality.

20250117-actors-pubsub.md
```diff
@@ -0,0 +1,140 @@
...
+
+The client response is similarly the same format as [today](https://github.com/dapr/dapr/blob/8cefe6a312867d8517a750d2dc300e10821bdf3a/dapr/proto/runtime/v1/appcallback.proto#L136).
+
+### Non-Actor Publishing
+
+Non-actor clients are able to publish messages to actor subscribers via a new `PublishActorEvent` RPC.
+This RPC is close to the existing `PublishEvent` rpc, accept that it takes an `actor_type` field rather than a `pubsub_name` field.
+This is because their is 1. only one pubsub associated with actors in that namespace, and 2. the fully qualified topic name is derived via the actor type.
+
+Publishing messages to actors will not invoke, and therefore initialize the existence of an actor, since an actor needs to be subscribing to a topic to receive messages.
```
theperm, line 110: I can see a requirement to also initialise an actor - could we have a flag that enables initialisation of an actor first.
The use case here is where the topic name is/contains the ID of the actor so the message is only destined for consumption for that actor id. 
If no other actor is subscribed what is the delivery semantics? will it remained queued waiting for the the actor to initialise, subscribe and receive? or does it go straight on the dead letter q. This might also depend on how the the underlying pubsub is configured and delivery semantics At-most-once or at-least-once and i persistent ques are involved? for example using NATS there is much more control over delivery.

20250117-actors-pubsub.md
```diff
@@ -0,0 +1,140 @@
...
+## Implementation Details
+
+### Actor PubSub Registration
+
+A single PubSub component will be designated as the actor PubSub, similar to how an actor state store is configured.
+The metadata configuration will follow the existing pattern:
+
+```yaml
+metadata:
+  - name: actorpubsub
```
artursouza, line 27: What is the advantage to have a single pubsub for actors vs allow it to be configured when registering the actor type, so different actor types can run with different pubsubs?

20250117-actors-pubsub.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+Only one PubSub component can be marked as the actor PubSub.
+
+### Actor Subscriptions
+
+Actors will be able to subscribe to messages within their type for a particular topic.
+Subscriptions will be dynamically declared, and messages for that topic will be routed through the established bi-directional gRPC stream.
+
+Upon deactivation, either during shutdown or actor rebalancing, the subscription to all topics by the actor will be removed.
```
artursouza, line 38: It is not clear how rebalancing will work with the subscription. If the subscription is per actor type, how do you handle redirecting the message to the right actor instance?

Also, how to handle subscription to all actorIds (fanout) vs to a single actor ID?

See https://www.youtube.com/watch?v=8Aj1WUzVvGs&t=1738s

20250117-actors-pubsub.md
```diff
@@ -0,0 +1,140 @@
...
+
+```yaml
+metadata:
+  - name: actorpubsub
+    value: "true"
+```
+
+Only one PubSub component can be marked as the actor PubSub.
+
+### Actor Subscriptions
```
artursouza, line 33: Maybe we don't need a special subscription. If you add the actor type and id in the cloudevent payload, routing can be done automatically from sidecar without the need of a new API.

artursouza (changes requested):

20250117-actors-pubsub.md
JoshVanL: > This may not be as bad, but we do have generally a large number of pods and so the mapping tables could get large. 

The same way this is handled with proxying existing requests today.

JoshVanL:
> > This feature should be implemented after the actor bi-directional streaming feature is completed.
> > The bi-directional stream will serve as the transport mechanism for delivering messages to actors.
> 
> Is there a proposal for this feature? As it could be considered alongside the desire for true async message passing (inbox / outbox) between Actors, as I outlined here [dapr/dapr#501 (comment)](https://github.com/dapr/dapr/issues/501#issuecomment-1689787965)

@olitomlinson the proposal for the bi-directional stream is here.

20250117-actors-pubsub.md
artursouza: Maybe a diagram can help explain this better. Each sidecar instance will read from the message broker, so will they compete for messages and redirect to the right sidecar for each actor id?

JoshVanL:
Thank for all the feedback @olitomlinson @yaron2 @theperm @artursouza @WhitWaldo !

Is seems to me to be the case that (at least the first iteration) should instead cover the use case of actors implicitly subscribing to a single actor type topic, where messages are published to an explicit actor ID which are invoked on a defined method. These messages will spawn that actor ID if it does not already exist.

Instead of updating this proposal PR, I thought it best to close this one and open a new one- I think this is better for posterity and review. Please don't think I'm ignoring your great feedback!

The new proposal over here: https://github.com/dapr/proposals/pull/74

PR #75, opened 2025-03-22
Alternative proposal for Actor PubSub implementation

WhitWaldo:
None of the comments have been addressed in the various reviews on the [existing Actors PubSub proposal](https://github.com/dapr/proposals/pull/74) nor has any discussion occurred on the thread I created in the maintainers channel on Discord.

But as there is voting occurring on the existing proposal, I'm making a last-ditch effort to suggest ditching that proposal in favor of an alternative that I think is more broadly applicable to what people want to get out of Actors and PubSub, repurposes what's already available and used in Dapr and avoids having additional "one component for all occasions" dependencies.

Fundamentally, I take what we already use for Dapr PubSub today (declarative, programmatic and streaming subscriptions) and extend it to Actors. Contrary to the existing proposal which changes how anyone would have to send events in to use them in actors, my proposal allows actors to subscribe to any existing PubSub component without changing how the events come in and only changing how actors themselves subscribe. In other words, mine is a drop-in change that anyone can use on day-one with existing PubSub without needing anything more than annotating actor methods with corresponding routing (for declarative or programmatic subscriptions).

I welcome any and all feedback!

### DIFF
--- a/20250322-actors-pubsub.md
+++ b/20250322-actors-pubsub.md
@@ -355,7 +355,7 @@ subscriptions at runtime is a lightweight operation as these subscriptions may b
 - All filtering should occur on the runtime to avoid activating actors unnecessarily.
 - If a registered sink activates an actor that hasn't previously been activated, it should activate it like any other inbound
   request would.
-- PubSub invocations should follow the turn-based limitations inherent to Daor Actors already. As such, this may require
+- PubSub invocations should follow the turn-based limitations inherent to Dapr Actors already. As such, this may require
   the creation of an inbox of sorts for each actor to handle queued messages pending successful acknowledgement by the
   actor.
 
--- a/20250322-actors-pubsub.md
+++ b/20250322-actors-pubsub.md
@@ -246,8 +246,8 @@ service Dapr {
 // The message containing the details for subscribing an actor to a topic via streaming or otherwise
 message SubscribeActorEventRequestAlpha1 {
     oneof subscribe_topic_events_request_type {
-        SubscribeActorTopicEventsRequestInitialApha1 initial_request = 1;
-        SubscribeActorTopicEventsRequestProcessedAlpah1 event_processed = 2;        
+        SubscribeActorTopicEventsRequestInitialAlpha1 initial_request = 1;
+        SubscribeActorTopicEventsRequestProcessedAlpha1 event_processed = 2;        
     }
 }
 
--- a/20250322-actors-pubsub.md
+++ b/20250322-actors-pubsub.md
@@ -156,7 +156,7 @@ public async Task<Order> ProcessOrders(Order order)
 
 Like the declarative subscription, this identifies a route on the actor that:
 - Maps to the `pubsub` component
-- Subscribes to the `orders` topic or queue nme
+- Subscribes to the `orders` topic or queue name
 - Handles subscriptions using the `ProcessOrders` method implemented on the actor
 
 When used without a paired declarative subscription, these routes are identified exclusively at startup and are **not**
--- a/20250322-actors-pubsub.md
+++ b/20250322-actors-pubsub.md
@@ -1,0 +1,367 @@
+﻿# Alternate Actor PubSub Implementation Proposal
+- Author(s): @joshvanl, @whitwaldo
+
+## Overview
+This proposal introduces PubSub capabilities for Dapr Actors, enabling actor instances to subscribe to routed messages
+directly regardless of what provided the messages. Actors will be spawned, if they are not already, when they receive
+an actor PubSub message that matches their registered routing filters. Because actors are virtual, individual actor
+instances can subscribe to a PubSub topic or queue, but they will only receive the messages sent through it following
+that subscription.
+
+This proposal differs from the [original Actors PubSub proposal](https://github.com/dapr/proposals/pull/74) in the
+following ways:
+- There is no designated PubSub component in actors
+- Declared subscriptions to a whole actor type will only push messages to instances activated before that point
+- Individual actor instances can set up their own subscriptions independent of type
+- A given actor instance can subscribe to more than one PubSub component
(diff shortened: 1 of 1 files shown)


WhitWaldo:
Pinging the following who have expressed interest in this proposal for visibility:
@JoshVanL @yaron2 @msfussell @elena-kolevska @olitomlinson @nelson-parente @cicoyle @theperm

20250322-actors-pubsub.md
```diff
@@ -0,0 +1,367 @@
...
+    // Required.
+    string topic_name = 2;
+    // Registers the sink to deliver the messages to.
+    // Required.    
+    repeated oneof subscribe_actor_event_request_filter_type {
+      // Registers a filtered subscription with a CEL expression that must evaluate as true
+      SubscribeActorTopicEventsRequestFilterAlpha1
+      // Registers a non-filtered subscription
+      SubscribeActorTopicSinkAlpha1 
+    } = 6;
```
JoshVanL, line 269: What's the benefit of using a `oneof` here, rather than having `expression` as an `optional`?

20250322-actors-pubsub.md
```diff
@@ -0,0 +1,367 @@
...
+To be clear, this means that the original proposal's "actorpubsub" annotation on a single pubsub component would **NOT**
+apply here as there would be no "central" pubsub component used. Rather, this proposal is a broadening of the existing
+PubSub functionality paired specifically with existing Actor use cases.
+
+### Actor Type Subscription
+Like the original proposal, on actor type registration, if the specified PubSub component is available, the Dapr runtime
+will subscribe to the specified queue/topic indicated. This topic string will take the format of:
+
+```
+dapr.actors||$namespace||$actorType
```
antontroshin, line 62: Does this mean that a `namespace` is required for all subscriptions?
Having `namespace` as optional can be both beneficial and detrimental to some systems, depends on use case of course.
Or it is assumed that both publishers and subscribers operate in the same namespace and actor type pubsub messages will be routed only within the originating namespace?

20250322-actors-pubsub.md
WhitWaldo: No real benefit - just copied the original approach you had.

20250322-actors-pubsub.md
WhitWaldo: I simply followed the text from the original proposal here. I'd be inclined towards making it optional to favor more scenarios than fewer.

WhitWaldo:
Following the maintainer call on 3/24, @JoshVanL and I talked through several aspects of this proposal and I wanted to write them down for posterity:

### Shift from existing PubSub approach - single Dapr consumer instead of per subscriber
Where it makes sense for the PubSub broker to set up a new consumer for each subscription today, this isn't practical with PubSub actors as offhand, I cannot think of a PubSub provider that supports more than 32 consumer groups, much less an infinite amount (e.g. the potential identifier space for actor IDs). Rather, I propose that this be handled differently in that the Dapr runtime itself be the sole consumer of inbound messages and that it itself is responsible for attempting to locate a subscribed message consumer. If a consumer is identified and the message conveyed, the message should be acknowledged back to the PubSub component (and retries to subscribers handled per resiliency policies). If no consumer is found, the messages should be either retried or dropped (per policy).

### Subscription Types
We agreed that the most valuable subscription type here is streaming. I expressed support for the notion that like Orleans, the concept of a PubSub subscription should be one where they're potentially either long-standing or brief and favor dynamic creation/deletion. To that end, I'd be entirely OK with dropping declarative subscriptions from the initial release. While it could be nice to have programmatic subscriptions (applied only at service startup and applicable only to previously activated actor instances), if the initial version only allows for streaming subscriptions, I think that'd be a great place to start.

Existing declarative PubSub cannot point to actors for lack of identifying actor names and instance IDs.

### Filter Type
Today, we favor Google's CEL approach for handling event filtering, but it's inconsistent with the approach suggested in CloudEvents. While I modified the CloudEvent schema to support this CEL approach, we might want to take advantage of the opportunity here of a few feature to instead wholeheartedly embrace conformance with CloudEvents instead of CEL.

### Subscription de-duplication
Because I propose that the filter be assigned on the runtime (to avoid unnecessary actor activation), in the short term this introduces the opportunity to do subscription de-duplication on a per-actor ID basis. Reiterating from the proposal, I think it's very possible that a single actor instance may have more than one subscription to the same PubSub component and topic (and potentially with the same filter across multiple methods) so there's an opportunity to at least put all these in the same record.

That said, I would eventually (as part of a separate proposal) support another Orleans capability of rewinding subscriptions.  When a subscription is created, it naturally receives only the next message received, but if the underlying provider supports sequence tokens (e.g. like [Azure Event Hub](https://learn.microsoft.com/en-us/azure/event-hubs/event-hubs-features#stream-offsets)), these can be passed back by Dapr from the subscriber to pull earlier messages prior to subscription. This might have an impact on runtime-side de-duplication, so it's something to consider in advance.


## Unmentioned on Call
There are a few additional tweaks I'll add to the PR itself, but since I'm thinking about it:
- Should have a means of getting an existing subscription details (even if just by ID)
- Should have a means of listing existing subscriptions for a given actor ID
- Shift to supporting CloudEvents [1.0.2 standard](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md)
This is sort of covered above in the note about CEL vs CloudEvent filtering, but worth repeating. Today, Dapr assumes that if the content type isn't provided, it must be `text/plain` and rather assumes that the data is JSON or similar for filtering. I would instead propose that instead of having our own funky hybrid of this standard, we take the opportunity to fully support CloudEvent including support for binary or text payloads on events. and adopt the [subscriptions spec](https://github.com/cloudevents/spec/blob/main/subscriptions/spec.md) for this PubSub actor functionality.

WhitWaldo:
Closing in favor of #78 since I put this on the wrong local branch and can't change it

PR #76, opened 2025-03-24
Multi-App Workflows

JoshVanL:


olitomlinson:
+1 for this proposal...

...as long as security boundaries are not broken. If they are broken, then a viable permission model is accompanied to allow a developer to specify which external Apps are allowed to call management operations on the Host Workflow Application*, which defaults to secure-by-design / zero trust)

_Host Workflow Application is the application where the Workflow is registered._

WhitWaldo:
My only note here would be that if the specified app ID cannot be found, or the type cannot be found on the app ID or just generally something happens that is unexpected (e.g. security boundary violation), it'd be nice to have some context in the error returned by the runtime to the SDK so it can be handled and addressed programmatically and clearly instead of requiring an Easter egg hunt by the developer after the fact to figure out why it broke (e.g. today it just returns a 500 for either of the first two scenarios) - recent request for this [on Discord](https://discord.com/channels/778680217417809931/886835700769644595/1363855666258907377).

yaron2:
+1 binding

yaron2 (approved):

olitomlinson:
It saddens me to see this proposal merged without receiving _any_ feedback to what I believe are valid questions around how this impacts the existing workflows security model.

olitomlinson:
I forgot to remind about this security consideration here @cicoyle - has this being catered for in the implementation?

https://github.com/dapr/dapr/issues/8578#issuecomment-2732603999

 **Potential areas of concern - Discoverability of Activities across App-boundaries**

 - Apps are the security boundary of workflows and their registered Activities.
 - **Allowing Activities to be used from outside of the App boundary would create a security hole.**
   - Therefore we must have a way to mark each Activity as discoverable, so that it can be discovered and used by one or more other Apps in the mesh. Dapr already has an [ACL](https://docs.dapr.io/operations/configuration/invoke-allowlist/) concept for Service Invoke, can we extend the ACL with Workflows in the same/similar manner?

PR #79, opened 2025-03-25
Workflow Checkpointing Proposal

WhitWaldo:
On the maintainers' call on 3/24/25, @JoshVanL proposed an idea he'd been thinking about regarding resuscitating workflows that were in a terminal state and shared some background around the origins of the idea. While thinking about what this might look like, it occurred to me that there was potentially a simpler approach that accommodated the same conclusion, but which might fit better in the existing Dapr Workflows paradigm, so I've typed this up to get broader feedback.

### DIFF
--- a/20250325-BS-workflow-checkpoints.md
+++ b/20250325-BS-workflow-checkpoints.md
@@ -1,0 +1,127 @@
+﻿# Workflow Checkpointing
+- Author: @whitwaldo
+
+## Overview
+This design proposal augments the implementation for Workflows to support the creation of sub-workflows from 
+"checkpoints" within existing workflows.
+
+## Background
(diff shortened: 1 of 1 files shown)


jjcollinge:
> I strongly recommend that you sync up with @cgillum as he has extremely valuable prior art here with the **Rewind API** in Azure Durable Functions, which never made it to GA because of challenges.
> 
> Personally, I would like to see this solved at the DurableTask level and I would also like to see the proposal considered along side versioning and patching (similar to Temporals approach) to see which has the most immediate and impactful value to customers / bang-for-engineering-buck

Linking https://github.com/Azure/azure-functions-durable-extension/discussions/1828

PR #81, opened 2025-04-14
Proposal to change release feature planning

WhitWaldo:
Thank you for giving it a read and for your consideration!

### DIFF
--- a/20250412-P-feature-planning.md
+++ b/20250412-P-feature-planning.md
@@ -1,0 +1,182 @@
+﻿# Improving the Dapr Feature Planning
+
+* Author(s): Whit Waldo (@whitwaldo)
+* State: Proposed
+* Date: 4/12/2025
+
+## Overview
+This proposal aims to formalize a new approach to planning releases, enhancing feature eligibility criteria, and
+bringing transparency, clarity and definitiveness to the current process.
+
+### Current Process
+After releasing a new Dapr version, maintainers and contributors meet weekly to discuss high-level goals for the
+next release. We review what happened in the previous release interval, identify improvement opportunities (process,
+tooling, new capabilities) and draft proposals reflecting these discussions over the following weeks.
+
+These proposals are debated and iterated upon until consensus is reached. Developers and assigned tasks,
+POCs are built, tests are added, documentation is produced, SDKs are updated, and features may be highlighted in a
+community call before being released.
+
+### Shortcomings
+The current process has several shortcomings:
+- **Pending Proposals:** Many proposals in dapr/proposals have been iterated on for months or years by numerous 
+contributors but remain inactive due to lack of advocacy. Despite broad community interest across a long period, these
+proposals are often overlooked.
+- **New Proposals:** In contrast, new proposals benefit from only a week of feedback from a limited number
+of contributors. This rushed process often results in features that may not align with larger project goals or
+provide lasting positive impact.
+- **Artificial Urgency:** Setting tentative release dates upfront and selecting priorities before designing them creates
+unnecessary urgency. This rush to consensus risks half-baked ideas being implemented and potentially impact
+credibility around release timelines as issues that might have been caught during the design process are identified
+and iterated on on-the-fly in a rush to get the feature completed.
+
+The current release cycle structure is ineffective as it forces us to race potentially incomplete ideas out the door.
+
+## Proposal
+I would like to propose several changes to our release process starting immediately.
+
+### Develop Proposals Between Releases
+We launched v1.15 on February 26th. Each of the current goals of this release span proposals introduced largely around
+and since then:
+
+| Proposal | Introduced At                                     | Proposed By |
+| -- |---------------------------------------------------| -- |
+|Actors: Bi-directional streaming | February 17                                       | Josh van Leeuwen |
+| Actors: PubSub | February 26, alternative proposal raised March 25 | Josh van Leeuwen, alternative by Whit Waldo |
+| Multi-App Workflows | March 24                                          | Josh van Leeuwen |
+| Event Streaming | March 24                                          | Cassie Coyle |
+| Workflow Re-runs | March 25, alternative proposal raised April 6     | Whit Waldo, alternative by Josh van Leeuwen |
+
+Despite being raised no more than a couple of weeks ago, many of these proposals have already gained significant traction. 
+However, implementing well-designed flagship features based on a three-week ideation and design timeline is unrealistic
+and isn't the process taken with larger software groups. It shouldn't be done here either. Furthermost, most of these
+issues have minimal community presence outside of the most-active contributors themselves.
+
+#### Proposed Changes
+Now is the time for brainstorming big-picture ideas when contributors take breaks from coding or casual discussions
+spark new inspiration.
+
+Discord frequently features requests for smaller features that could collectively form larger improvements both in terms
+of quality of life and feature additions. Even if these authors don't write original proposals themselves, we should document
+these ideas (in a place the stalebot won't hide them) and solicit community feedback to gauge traction. Regular community
+and meteor calls should highlight these new ideas with balanced commentary from hosts and direct viewers to relevant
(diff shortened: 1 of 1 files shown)


20250412-P-feature-planning.md
```diff
@@ -0,0 +1,182 @@
...
+
+#### Proposed Changes
+Now is the time for brainstorming big-picture ideas when contributors take breaks from coding or casual discussions
+spark new inspiration.
+
+Discord frequently features requests for smaller features that could collectively form larger improvements both in terms
+of quality of life and feature additions. Even if these authors don't write original proposals themselves, we should document
+these ideas (in a place the stalebot won't hide them) and solicit community feedback to gauge traction. Regular community
+and meteor calls should highlight these new ideas with balanced commentary from hosts and direct viewers to relevant
+issues so they can chime in themselves.
```
olitomlinson, line 63: I think we can all agree it's important that we encourage end-users to self-report issues as GitHub issues where they don't already exist as GitHub Issues. If those issues do already exist, it is important that we encourage users to add their feedback and votes of support to those issues.

Discord is a fine place to have conversations and explore issues, but it **must** result in an output into GitHub Issues/proposals to be super effective as a means for building evidence and datapoints for future considerations.

20250412-P-feature-planning.md
```diff
@@ -0,0 +1,182 @@
...
+expertise and a willingness from unknown other parties. Instead, we can change the process itself on a shorter timeline,
+and I propose we do so.
+
+## Conclusion
+The proposed changes to the Dapr feature planning process aim to address the current shortcomings by fostering a more 
+inclusive, transparent, and well-considered approach. By extending the proposal development period, we ensure that 
+ideas are thoroughly vetted and refined, reducing the risk of rushed, incomplete implementations that may be effectively
+rendered effectively dead on arrival. Additionally, broadening the voting pool to include all maintainers will 
+democratize decision-making, ensuring that features not only relevant to a few, but to a broader group of maintainers based
+on their personal experiences and the diverse parts of the community they represent.
```
olitomlinson, line 174: This is certainly something that has been discussed in the STC as part of an objective to increase project diversity in 2025

I strongly agree that having more binding votes from a wider pool of dapr maintainers would in general benefit of the Open Source project. 

I also think that this move would potentially attract more SDK maintainers, as they now would have a greater say in the core direction of the project, rather than just the SDKs, which is a great selling point.

yaron2: This specifically would prove very detrimental to the project, and will go against how most (if not all) of CNCF projects assign binding votes, many of which go much more granular and only assign specific voting rights for sub-sections of the code in a given repository.

The reason for this is to allow subject matter experts to make decisions based on their deep, ongoing understanding of their area of ownership, after having worked and achieved a level of understanding that can allow one to weigh in on what often times are critical changes. Giving SDK maintainers the ability to vote on a codebase they are not necessarily familiar with, on a language they are not proficient in (A PHP maintainer voting on a Go codebase, for example) can result in decisions that would harm the project rather than help it. This goes the other way around, as well - If you open it up this way, you will get votes for SDK based decisions from developers who do not have the deep, ecosystem-wide understanding required to keep SDKs feeling natural and true to their language of choice.

Another fallout of allowing this would be creating bottlenecks in voting. When you have a much larger pool of votes, votes can hang sometimes indefinitely due to people being out, sick, on leave or just overwhelmed with other work. What then happens is people are pinged continuously to provide their vote, and often times a vote will be given in a rushed way to just "get this done" and not to unblock others. I know this well because we've been there in the past, and that hasn't worked well. I've also been privy to other CNCF project maintainers chats talking about the same experiences and lessons learned.

So, I am very much in favor of soliciting feedback from more people in the community and taking that feedback into account to increase the diversity of opinions, but that should not come down to indiscriminate voting rights which will have the ability to throw the entire project out of balance.

20250412-P-feature-planning.md
```diff
@@ -0,0 +1,182 @@
...
+
+Designing large-scale features without time pressure soliciting feedback from less-invested contributors can help ensure
+that new capabilities are well-thought-out and either fit into our larger vision or require further refinement.
+
+### Release Prioritization
+Currently, we ideate high-level goals on maintainer and design calls following a release and wait for proposals to be
+submitted, and then largely just move forward with them.
+
+#### Proposed Changes
+I propose a substantial change to our process. Barring hotfixes for intermediate patch releases, only ideas backed by an
```
yaron2, line 73: I like this idea, just feeling that the "two months" period should have some level of justification for being that amount of time and not another

20250412-P-feature-planning.md
olitomlinson: I understand your concerns, specifically about people who don't have Go experience being in a position of authority over the runtime. It's a fair point.

Putting Go skillset aside, how do non-maintainers of the runtime (like Whit, for example), take some equity in the runtime from a **Product** perspective? To me, It's becoming clearer that as the user base of Dapr grows, we need to add more product-led thinking into the roadmapping and prioritisation aspects of the project.

As it currently stands, for proven maintainers (like Whit), there should be away to achieve some **equity** in the OS product vision/direction of the runtime, like a maintainer would have.

_Thinking of ideas..._

Would it be entirely unreasonable to create a new non-maintainer role (more akin to a Technical Product Owner / Product Manager role) in the runtime which affords n equitable vote over product-related decisions rather than internal implementation details? 

Maybe such a role is only attainable via a vote by the STC. e.g. "Once an SDK maintainer has been committed for 12 months (plus other criteria), they are eligible for applying for a Technical Product Owner role on the runtime" or something akin.

I propose that demonstrating that there is a pathway to allow **proven** people an opportunity to influence the project with a would be a testament to Daprs commitment to diversity of opinion.

20250412-P-feature-planning.md
WhitWaldo: While we aim for four releases a year, we've been hitting three releases a year, meaning each release is taking on average 4 months or so to complete. So I just picked 2 months as that's half of a 4-month release cycle and feels like a sufficiently long enough period for a "last-minute" idea to start baking and getting public feedback, project alignment, design iteration, etc.

20250412-P-feature-planning.md
yaron2: > As it currently stands, for proven maintainers (like Whit), there should be away to achieve some equity in the OS product vision/direction of the runtime, like a maintainer would have.

Strong disagree. In the CNCF, and even more so in Apache, projects are assigned maintainers (committers in Apache) on a per-project basis and maintainers are scoped with deciding the technical direction of the project, including everything you'd put in product-level decisions. 

> Would it be entirely unreasonable to create a new non-maintainer role (more akin to a Technical Product Owner / Product Manager role) in the runtime which affords n equitable vote over product-related decisions rather than internal implementation details?

There are no such roles that I'm aware in the CNCF or elsewhere, however if you can find an existing model somewhere that works, I think it'd be great because then we'd at least have the ability to audit something that works (or doesn't) and make an informed decision instead of coming up with a bespoke model that will make Dapr stand out from its peers. Also, there is the STC which allows every member, regardless of Go skills, to bring up any high level changes in the project which align more with product level thinking (recent examples: Dapr Agents, new APIs etc.). This is generally the established way for non-technical people to contribute to and vote on major decisions related to the project. This is less about technical implementation details of the runtime, which, as established, should be left to the maintainers of the given project as the subject matter experts.

> Putting Go skillset aside, how do non-maintainers of the runtime (like Whit, for example), take some equity in the runtime from a Product perspective?

Projects in the CNCF do not have a separate product function or element. The established and widely used maintainer model is based solely on merit and equal opportunity, which allows absolutely anyone who invests the time and effort required to make informed decisions to become a stakeholder and decision maker in the project and be voted as a maintainer.

WhitWaldo:
I have updated the proposal to limit it only to those aspects that have support and to remove the governance changes to a separate proposal for further discussion.

PR #83, opened 2025-04-29
Scheduler Broadcast Job Type

JoshVanL:


20250429-RS-scheduler-broadcast-job-type.md
```diff
@@ -0,0 +1,320 @@
...
+import "google/protobuf/any.proto";
+
+option go_package = "github.com/dapr/dapr/pkg/proto/scheduler/v1;scheduler";
+
+// BroadcastJobEventType is the type of the event why this job is being
+// broadcast.
+enum BroadcastJobEventType {
+  // BROADCAST_JOB_EVENT_PUT is the event type for a job being created or
+  // updated.
+  BROADCAST_JOB_EVENT_PUT = 0;
```
WhitWaldo, line 116: I'd propose that we favor immutability from the get-go and treat the job event registrations as immutable, requiring a delete and recreate instead of allowing updates.

The use of some sort of one-shot `AddOrReplace` method would be useful here so the operation can be done from the caller in one operation (avoid race or inconsistency conditions).

20250429-RS-scheduler-broadcast-job-type.md
```diff
@@ -0,0 +1,320 @@
...
+
+The actor runtime will receive events about the job;
+- on put events the actor runtime will create the underlying subscription, or replace one that is already existing.
+- on delete events the actor runtime will delete the subscription.
+- there should not be any trigger events.
+
+The Actor Subscription broadcast job will take the following form.
+Notice the schedule is set to infinite as the Job should never be triggered.
+The name of the Job object can be any application defined name.
+This job type will return an `already exists` error if a PUT is attempted on an existing Actor Subscription job.
```
WhitWaldo, line 198: Good - this aligns with my earlier comment (which should probably be changed then to remove "update").

20250429-RS-scheduler-broadcast-job-type.md
```diff
@@ -0,0 +1,320 @@
...
+  encoded invocation request.
+- on delete events the actor runtime will delete the durable actor ID.
+- there should not be any trigger events.
+
+The Durable Actor ID broadcast job will take the following form.
+Notice that the due time may be set to infinite (the actor ID can only be deleted via a delete event), or an optional TTL can be set to have automatic deletion.
+Since the job will be created abstractedly from the application by the runtime, the name will be generated with the following form; `__internal__$actor-type__$actor-id`.
+This job type will return an `already exists` error if a PUT is attempted on an existing durable Actor ID.
+The data field of the job will include a `common.InvokeRequest` proto message, detailing how the actor will be invoked.
+The HTTP Verb of the invocation request defined will be ignored, in favour of always using a PUT or DELETE.
```
WhitWaldo, line 265: Minor nit: I would prefer POST just because PUT implies an update, but POST almost always implies creating something new.

20250429-RS-scheduler-broadcast-job-type.md
```diff
@@ -0,0 +1,320 @@
...
+It is also the case that these messages need a reliable way to be broadcast to all the relevant hosts.
+
+While these two feature do have access to reliable durable storage in the form of the Actor State Store, implementing these require durable service discovery and co-ordination _at the runtime level_.
+Not to mention that storage is often not available outside of the context of Actors.
+Implementing these at the runtime level would be brittle as the co-ordination mechanism itself would require significant machinery.
+The Scheduler service already has application typed context service discovery and connectivity with durable storage access, so is the perfect place to implement this functionality.
+
+Users today are required to use "normal" Scheduler jobs to ensure message durable execution, and implement their own broadcast or message routing functionality.
+Typically creating a job which has an infinite schedule that triggers at a fixed small interval.
+Moving this functionality as a first concept to Dapr improves performance and improves the developer experience in creating complex distributed systems.
```
nelson-parente, line 30: How does this improve performance compared to "normal" Scheduler jobs to ensure message durable execution"?

20250429-RS-scheduler-broadcast-job-type.md
```diff
@@ -0,0 +1,320 @@
...
+
+This proposal describes adding a new Scheduler Job type - `Broadcast`.
+This job type is designed to leverage the durability of Scheduler and its reliable connection to all daprd runtimes to broadcast objects to all relevant clients to allow for durable message broadcasting.
+The initial design focuses on two initial use cases, `actor pubsub` and `durable actor IDs`, but this job type's functionality can be extended to other domains in future.
+
+## Related Items
+
+https://github.com/dapr/proposals/pull/78
+
+## Background
```
nelson-parente, line 16: nit: we talk about "durable" a lot, we could define that in the background

20250429-RS-scheduler-broadcast-job-type.md
```diff
@@ -0,0 +1,320 @@
...
+A client failing to return the response of a trigger should be considered as a failed connection.
+In any case, to account for future use cases, the trigger response codes should considered a logical AND whereby failed calls will be retired on each failed client call according to the Jobs Failure Policy.
+
+Broadcast jobs can either expire according to its schedule, or be deleted by a client.
+
+Broadcast Jobs will continue to adhere to the staging queue, like existing Jobs today.
+
+#### Keep Alive
+
+The gRPC keep alive options for both the Scheduler server and daprd clients need to be updated so that they are more aggressive.
```
nelson-parente, line 65: why does this need to change?

nelson-parente (commented):
🚀

20250429-RS-scheduler-broadcast-job-type.md
WhitWaldo: My take is that today we have "normal" Scheduler jobs for jobs and reminders. This gives us a mechanism to extend the concept to durable scheduled activities (as necessary for actor pub/sub and likely other use cases). I don't know it necessarily improves performance so much as abstract away the downstream purpose from the scheduler itself.

20250429-RS-scheduler-broadcast-job-type.md
WhitWaldo: Just about everything is stateless with regards to Dapr. Components are registered with the runtime, but if the runtime shuts down and a component removed, it doesn't have any sort of memory that it ever was. Once a message is brokered through the runtime for PubSub, there's no history to recall of it. Arguably, this is part of the case for the [event streaming proposal](https://github.com/dapr/proposals/pull/77) is that after a workflow lifecycle event occurs, that's it, it's gone - the runtime isn't persisting it anywhere. When we look at streaming pubsub subscriptions, the application registers the gRPC channel with daprd for a given component and topic, but if the application crashes, the channel is closed and that's the end of it. 

The difference with Actor PubSub is that when an actor subscribes to a PubSub topic, this time, we need information about that subscription to persist somewhere because if the actor is unloaded, that subscription still needs to be saved somewhere. The distributed scheduler is an excellent place for that because it can hold a little bit of state with each scheduled job, it's already architected to be quite scalable and it's very performant. So, the durability here is in regards to where we can keep this little bit of subscription data where it'll persist between actor invocations and even sidecar restarts (unlike nearly anything else in Dapr, except Jobs and Reminders).

20250429-RS-scheduler-broadcast-job-type.md
JoshVanL: It improves performance because today, one would have to rely on regular triggering of a job to signal a "keep alive" for the object to exist, without any easy way to signal a message has been deleted. Each trigger needs a round trip through the scheduler queue and call to each client over the network.

 This broadcast job type achieves CRUD message behaviour with the least number of Scheduler queue events and network calls .

20250429-RS-scheduler-broadcast-job-type.md
JoshVanL: ^

20250429-RS-scheduler-broadcast-job-type.md
JoshVanL: To ensure we have better settings to ensure the health of each stream to Scheduler. This was written more as a TODO for myself to improve the integrity of the client connection pool state. Realistically, this shouldn't have any behavioral changes to the API.

yaron2 (approved):

PR #84, opened 2025-05-24
Add Sentry JWT and OIDC Proposal

jjcollinge:
Signed-off-by: Jonathan Collinge <jonathancollinge@live.com>

### DIFF
--- a/20250524-R-sentry-jwt-oidc.md
+++ b/20250524-R-sentry-jwt-oidc.md
@@ -1,0 +1,88 @@
+# Sentry JWT and OIDC Support for Workload Identity
+
+* Author(s): Joni Collige
+* State: Implemented
+* Updated: 2025-05-24
+
+## Overview
+
+This proposal adds support for configuring Dapr Sentry to issue JWTs as an additional credential alongside the existing x.509 certificate. The JWT will be signed with a configurable key and algorithm, and a JWKS endpoint will be provided for trust distribution. Optionally, Sentry can host a partial OIDC discovery server to allow third-party identity providers to federate trust using the `.well-known/openid-configuration` and `jwks.json` endpoints. The JWT will be returned in the existing `SignCertificate` response, requiring no API changes. This enables Dapr workloads to use JWTs for federated cloud credentials and infrastructure access.
+
+* Areas affected: Dapr Sentry, Dapr runtime, Dapr components
+* Proposed: Add JWT and OIDC support to Sentry, with configuration for signing keys, algorithms, JWKS, and OIDC server options
+
+## Background
+
+Dapr currently issues x.509 SVIDs for workload identity using SPIFFE. Many cloud providers and third-party systems only support federated authentication using JWTs and OIDC discovery endpoints.
+Supporting JWTs in Sentry enables Dapr workloads to:
+- Use JWTs as credentials for identity provider federation
+
+This proposal addresses the need for flexible, standards-based workload identity in Dapr, improving security and interoperability for cloud-native and hybrid environments.
+
+## Related Items
+
+### Related proposals
+- [20231024-CIR-trust-distribution.md](20231024-CIR-trust-distribution.md): Trust bundle and CA distribution
+
+### Related issues
+
+## Expectations and alternatives
+
+* In scope:
+  - Sentry issues JWTs as workload credentials, signed with configurable key/algorithm
+  - JWKS endpoint for trust distribution
+  - Optional OIDC discovery server
+  - Audience claim configurable per app ID (default: trust domain)
+  - JWT rotation mechanism
+* Not in scope:
+  - Full OIDC authorization flows (only discovery and JWKS endpoints)
+  - Custom claims beyond audience, issuer, and standard fields
+* Alternatives considered:
+  - Only supporting x.509: limits interoperability with cloud and OIDC systems
+  - External JWT/OIDC proxy: adds operational complexity, not integrated with Sentry
+* Trade-offs:
+  - Adds complexity to Sentry configuration and deployment
+  - Exposes unauthenticated endpoints (user must secure as needed)
+* Advantages:
+  - Enables modern, federated identity scenarios
+  - No breaking changes to existing APIs
+* Disadvantages:
+  - More configuration required for Sentry
+  - Responsibility for endpoint security is on the user
+
+## Implementation Details
+
+### Design
+
+- Sentry configuration extended to accept JWT signing key, algorithm, and JWKS (or auto-generate if not provided)
+- JWT issued alongside x.509 SVID in `SignCertificate` response
+- Audience claim set from app ID configuration (default: trust domain)
+- JWKS endpoint served for public key distribution
+- Optional OIDC discovery server serves `.well-known/openid-configuration` and `jwks.json` endpoints
+- OIDC server exposed via Kubernetes Service; user responsible for external exposure and protection
+- JWT rotation uses same mechanism as x.509 rotation
+- Implementation extends existing SPIFFE package to support JWT-SVIDs
+- daprd runtime and components can use JWT for cloud federation via token provider
+
+### Feature lifecycle outline
+
+* Expectations: Feature is opt-in, backward compatible, and does not affect existing x.509 flows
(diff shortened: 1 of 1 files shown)


20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+- `--jwt-enabled` (bool): Enable JWT token issuance by Sentry
+- `--jwt-key-filename` (string): JWT signing key filename
+- `--jwks-filename` (string): JWKS (JSON Web Key Set) filename
+- `--jwt-issuer` (string): Issuer value for JWT tokens (no issuer if empty)
+- `--jwt-signing-algorithm` (string): Algorithm for JWT signing, must be supported by signing key
+- `--jwt-key-id` (string): Key ID (kid) for JWT signing
+
+**OIDC-related flags:**
+- `--oidc-http-port` (int): Port for the OIDC HTTP server (disabled if 0)
+- `--oidc-jwks-uri` (string): Custom URI for external JWKS access
```
JoshVanL, line 110: Default `nil`?

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+- `--oidc-path-prefix` (string): Path prefix for all OIDC HTTP endpoints
```
JoshVanL, line 111: “All”? Why is there more than one? If I understand this is for serving the oidc discovery endpoint?

Rather than prefix, please can we have explicit url path with sane well known defaults. Prefix style configuration is always confusing/finicky.

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+- `--oidc-domains` (string slice): Allowed domains for OIDC HTTP endpoint requests
```
JoshVanL, line 112: What does this do? Check the SNI headers for incoming well known http requests? 🤔

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+- `--oidc-tls-cert-file` (string): TLS certificate file for OIDC HTTP server (required if enabled)
+- `--oidc-tls-key-file` (string): TLS key file for OIDC HTTP server (required if enabled)
+- `--oidc-tls-insecure` (bool): Serve OIDC HTTP without TLS
```
JoshVanL, line 115:
```suggestion
- `--oidc-server-tls-cert-file` (string): TLS certificate file for OIDC HTTP server (required if enabled)
- `--oidc-server-tls-key-file` (string): TLS key file for OIDC HTTP server (required if enabled)
- `--oidc-server-tls-enable` (bool): Serve OIDC HTTP with TLS
```

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+**JWT-related flags:**
+- `--jwt-enabled` (bool): Enable JWT token issuance by Sentry
+- `--jwt-key-filename` (string): JWT signing key filename
+- `--jwks-filename` (string): JWKS (JSON Web Key Set) filename
+- `--jwt-issuer` (string): Issuer value for JWT tokens (no issuer if empty)
+- `--jwt-signing-algorithm` (string): Algorithm for JWT signing, must be supported by signing key
+- `--jwt-key-id` (string): Key ID (kid) for JWT signing
+
+**OIDC-related flags:**
+- `--oidc-http-port` (int): Port for the OIDC HTTP server (disabled if 0)
```
JoshVanL, line 109:
```suggestion
- `--oidc-server-listen-port` (int): Port for the OIDC HTTP server (disabled if 0)
```

Please also include a listen address flag like here (which defaults to localhost).

https://github.com/dapr/dapr/blob/137da857234b079199764324490bf4165a17a150/cmd/daprd/options/options.go#L123

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+- Additional configuration required for Sentry
+- Endpoint security is the user's responsibility
+
+## Implementation Details
+
+### Design
+
+- Sentry configuration extended for JWT signing key, algorithm, JWKS (auto-generated if not provided)
+- JWT issued alongside x.509 SVID in `SignCertificate` response
+- Audience claim set from app ID configuration (default: trust domain)
```
JoshVanL, line 55: Please include Configuration CRD changes.

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+### Proposed Sentry Configuration Flags
+
+**JWT-related flags:**
+- `--jwt-enabled` (bool): Enable JWT token issuance by Sentry
+- `--jwt-key-filename` (string): JWT signing key filename
+- `--jwks-filename` (string): JWKS (JSON Web Key Set) filename
+- `--jwt-issuer` (string): Issuer value for JWT tokens (no issuer if empty)
```
JoshVanL, line 104: Surely this is always the Dapr control plane trust domain, and cannot be configured differently?

JoshVanL (changes requested):
Please can you add a bit about private key generation, I.e, sentry does it on first boot. If multiple sentry's boot at the same time, how is first write wins managed? Is it possible to have multiple public keys in the file path that are advertised to enable key rotation?

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+- `--jwt-signing-algorithm` (string): Algorithm for JWT signing, must be supported by signing key
+- `--jwt-key-id` (string): Key ID (kid) for JWT signing
```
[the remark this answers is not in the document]
jjcollinge: This is the kid of the signing key. It's in the token and used by clients to uniquely look up the key in the JWKS during validation. I can make this a thumbprint of the key by default but allow the user to override it.

20250524-R-sentry-jwt-oidc.md
jjcollinge: We currently server the discovery and issuer JWKS endpoints. To be compliant we also have to advertise the authorize endpoint via the discovery doc even though it's not implemented. This is for routing - you need the discovery document endpoints to match the route you used to expose it.

20250524-R-sentry-jwt-oidc.md
jjcollinge: No, this has to match the domain you are exposing your OIDC server on.

20250524-R-sentry-jwt-oidc.md
jjcollinge: TLS is terminated here, but because you're exposing an unauthenticated HTTP server this just allows you to restrict the hostname in the request to ensure the requests are coming via the expected route e.g. `oidc.myissuer.net` and not via some internal or alternative domain.

20250524-R-sentry-jwt-oidc.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+### Proposed Sentry Configuration Flags
+
+**JWT-related flags:**
+- `--jwt-enabled` (bool): Enable JWT token issuance by Sentry
+- `--jwt-key-filename` (string): JWT signing key filename
+- `--jwks-filename` (string): JWKS (JSON Web Key Set) filename
+- `--jwt-issuer` (string): Issuer value for JWT tokens (no issuer if empty)
+- `--jwt-signing-algorithm` (string): Algorithm for JWT signing, must be supported by signing key
```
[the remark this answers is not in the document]
jjcollinge: Added

jjcollinge:
> Please can you add a bit about private key generation, I.e, sentry does it on first boot. If multiple sentry's boot at the same time, how is first write wins managed? Is it possible to have multiple public keys in the file path that are advertised to enable key rotation?

Added a bit on "Propose Key Management". I'm just extending the current Sentry mechanism when generating the keys - I don't see any leadership election etc. so I'm assuming we just accept LWW currently anyway?

PR #86, opened 2025-06-11
Add Config to Disable Sidecar Requests to /dapr/config and /dapr/subscribe

tomflenner:
Create a new proposal that target to answer the need of https://github.com/dapr/dapr/issues/8224

### DIFF
--- a/20250611-IR-disable-dapr-init-config-and-subscribe-requests.md
+++ b/20250611-IR-disable-dapr-init-config-and-subscribe-requests.md
@@ -1,0 +1,114 @@
+# Add Config to Disable Sidecar Requests to /dapr/config and /dapr/subscribe
+
+* Author(s): Tom Flenner
+* State: Waiting for approval to start implementation
+* Updated: 20250611
+
+## Overview
+
+This proposal introduces a new configuration option that allows disabling automatic HTTP requests to `/dapr/config` and `/dapr/subscribe` during Dapr sidecar initialization.
+
+Areas affected:
+
+>/area runtime
+
+>/area docs
+
+What is being proposed:
+
+Based on Josh van Leeuwen thoughts :
+
+Add support for a new `--disable-init-endpoints` flag in `dapr run`, as well as a corresponding `dapr.io/disable-init-endpoints annotation`. This flag will accept a string slice of endpoints (e.g., `["config", "subscribe"]`) to disable the automatic loading behavior per sidecar instance.
+
+## Background
+
+Dapr makes automatic HTTP requests to `/dapr/config` and `/dapr/subscribe` endpoints during initialization. While this behavior is helpful for applications relying on programmatic configuration or subscriptions, it introduces several issues for applications that do not use these features:
+
+- Log Noise: Unnecessary requests produce warning logs, leading to confusion for developers who do not expect or understand these failures.
+
+- Unexpected Failures: Under certain network conditions or misconfigured routes, these automatic requests can block declarative subscriptions from being loaded correctly. This results in broken functionality even though the app only intends to use declarative configuration.
+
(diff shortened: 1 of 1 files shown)


WhitWaldo:
As the end result is seeing fewer logs, I would propose a few changes:
- Rather than putting a flag somewhere that can be easily forgotten about, if there are no configuration components loaded by the runtime, simply disable the configuration functionality altogether and don't log any of it

PubSub is a little more complicated because of the opportunity for programmatic endpoints, but:
- If there's no pubsub component loaded by the runtime, simply disable all pubsub functionality altogether and don't log any of it
- If there is a pubsub component loaded and the CLI flag/argument is specified, disable all pubsub programmatic functionality (e.g. the flag disables programmatic functionality necessitating the dapr/subscribe endpoint) and relevant logging. In other words, the CLI flag/argument disables the progammatic pubsub functionality, not necessarily the rest of PubSub, _if_ a pubsub component is provided.

Otherwise, this looks great!

WhitWaldo:
> To be sure that I understand correctly the last bullet point - we are not hitting the `/dapr/subscribe` but we are displaying a log that will help to see that programmatic subscription is not registered on runtime side ? Or you really want to see that there are some potential subscriptions that can be hidden due to that flag in this particular setup

I envision a scenario where someone has launched one of the many examples or quickstarts which registers a number of programmatic endpoints to trial Dapr's PubSub capability. They google something like "Dapr CLI" and Gemini, bless it, yields a CLI prompt that contains your programmatic PubSub disable flag. They blindly copy and paste it into their command prompt and cannot figure out why the PubSub example isn't working - none of the programmatic endpoints appear to do anything. Reason being - they're disabled by the flag.

So this would serve as a backup - if the runtime is generally disabling programmatic PubSub despite the presence of a PubSub component, check to see if it appears that there are PubSub topic endpoints registered.  If there aren't, don't mention anything about it in the logs - programmatic PubSub is disabled and that's the last of it. But if there _are_ any such endpoints, put one warning in the logs that notes that while programmatic PubSub has been disabled because of the flag, this might be a mistake because there are programmatic endpoints that were registered on the application.

Otherwise, I envision this being another scenario that becomes difficult to troubleshoot - why aren't programmatic PubSub endpoints working? Don't know - the logs don't say anything about PubSub. Why not? Because you added a flag to disable the programmatic endpoints. Help future us - if it looks like disabling the endpoint was a mistake, log it.

tomflenner:
> if the runtime is generally disabling programmatic PubSub despite the presence of a PubSub component, check to see if it appears that there are PubSub topic endpoints registered.

This is by having the runtime call `/dapr/subscribe` of the app right ? So we still introduce a http request from the runtime to the app in this particular scenario where it's probable that we do want programmatic PubSub.

If it's the case - I try to figure out if there is any real world scenario where we want to disable this init request while using PubSub component with declarative subscription only for example. In that one, we might dont want any noise that maybe ?

WhitWaldo:
> What if you're working with a .NET minimal API and using declarative subscriptions? In that case, calling `/dapr/subscribe` doesn't really make sense.
> 
> However, you'll still receive an HTTP call to that endpoint, which could result in unnecessary log entries - both in your app and in the Dapr runtime - possibly including warnings about programmatic subscriptions being disabled.
> 
> Is this the desired behavior?

Until the app has loaded, the runtime couldn't possibly know if there are any programmatic subscriptions set up or not and it would only know this by checking this endpoint. 

Yes, your way of having the flag just disable checks against that endpoint altogether would remove all logging whatsoever, but I personally would rather know from the logs that I possibly misconfigured something over reduced verbosity around checking the endpoint at startup.

Put another way, I'd rather typically address something like this with a source analyzer, but the application doesn't know that which is only available to the runtime. So that leaves the runtime as the only entity with the full picture, which means I would favor it checking "just in case" and logging about the inconsistency over not.

PR #88, opened 2025-08-23
Proposal: Binary Store Building Block

WhitWaldo:
Increasingly, while writing applications that use Dapr, I keep running into the need to persist data that's too large to reasonably store using Dapr often because it's too large and will exhaust the memory resources of the sidecar, though frequently because it's likely too large to store in a key/value store. 

It doesn't make a ton of sense to rely exclusively on bindings for this when that really just provides a Dapr-hosted alternative to the provider's SDK for something that we should increasingly have broad provider support for. Object and blob stores are really overloaded terms representing all manner of things depending on provider for which I think there's a fine opportunity to tackle in the future - this proposal isn't that.

Here, I propose an API devoid of List and even Metadata operations so it can accommodate the broadest of possible storage providers and instead suggest that we increasingly lean on the SDKs to provide the state management instead of putting all that weight on the runtime and the components. It's a slim implementation that should be pretty easily added, but which would provide immediate benefits for popular Dapr features: Workflows and the new Agentic operations come to mind, but it would be beneficial for Actor and Cryptographic operations as well.

I look forward to your feedback!

### DIFF
--- a/20250823-BCS-file-store.md
+++ b/20250823-BCS-file-store.md
@@ -295,8 +295,17 @@ This is intended as a provider-agnostic, lean building block that provides a sim
 in a streaming fashion that's generally too large for a key/value store.
 
 I leave it to the architects of future even more specialized state store building blocks to decide whether they want to
-support additional features beyond what's provided beyond this narrow implementation. 
+support additional features beyond what's provided beyond this narrow implementation.
 
+### What if we called it `BinaryStore` instead of `FileStore`? (thanks @olitomlinson!)
+Love it. By calling out that it's a file store, I'm simply trying to avoid any confusion that a developer can pass some
+arbitrary in-memory representation and trust that some serialization magic will persist it. By requiring that it be a 
+file, we're ensuring that it's in a state that's persistable elsewhere. Here, I imagine that the SDK API would require 
+that the developer provide an array of bytes or a byte stream of already-serialized/encoded data as it needn't strictly
+be a location on their file system to a file.
+
+To that end then, I'm entirely ok with calling this a `BinaryStore` instead as it gives a similar level of 
+understanding if that makes more sense than `FileStore`.
 
 ## Completion Checklist
 - [ ] File Store Runtime Implementation
--- a/20250823-BCS-file-store.md
+++ b/20250823-BCS-file-store.md
@@ -1,0 +1,306 @@
+﻿# File storage building block
+- Author(s): @WhitWaldo
+- Significantly inspired by: @ItalyPaleAle (https://github.com/dapr/proposals/pull/18)
+- Updated August 23, 2025
+
+## Overview
+This is a design proposal for building a new specialized state store for managing named files in state providers. I
+intentionally use the work file as opposed to blob or object as to avoid any of the overloaded meaning associated with
+common object store or blob store providers. I think there's every possibility that it'll make sense to introduce
+a proper "blob store" or "object store" building block in the future. However, for now I think it's best to keep the
+scope of this proposal focused on files more generally and avoid the complexities of either of those specialty stores.
+Over time, the distinction between this and another Blob Store or Object Store blobk should be made clearer by the
+enhanced capabilities of either of those APIs (e.g. list by prefix or handling metadata), but this proposal should
+be understood to be a bare-boned implementation designed to strictly store and retrieve (potentially) large serialized
+files.
+
+Files are any contiguous sequences of unstructured data comprised of bytes that should be assumed to span lengths as 
+few as a handful of bytes to several gigabytes. Examples include images, videos, documents, database backups, 
+compressed archives and so on.
+
+Finally, while I also considered calling this the even more ambiguous `DataStore`, I want to be clear that this is 
+intended to persist data that's been serialized and not some store for arbitrary in-memory data that's not trivially 
+persisted externally. It should be called out in the documentation that this is not intended to be a replacement for 
+something like SMB or NFS and is just a general-purpose storage implementation that doesn't get into the Object/Blob 
+details.
+
+## Background
+Dapr currently implements a state store around a strictly key/value basis. This approach has led to component support
+across many different component providers and works for many common use cases. However, decisions around the design
+and implementation of the state store feature make it increasingly important to introduce a new generation of 
+special-purpose state management building blocks instead of having a one-size-fits-all approach to managing state.
+
+For example, files differ from values in key/value stores in that they can frequently span large amounts of data that
+isn't generally compatible with the likes of Redis or Etcd. Because they span such large amounts of data, it's frequently
+insufficient to simply store and retrieve them synchronously, mandating a stream-first approach to minimize
+resources needed by the runtime to transfer the data. The goal is to allow users to store and retrieve files in a 
+performant and scalable manner via a common runtime API and enable the Dapr SDKs to complement the implementation to
+support additional provider-specific features on a case by case basis. A possible example of this will be given below.
+
+## Component YAML
+This component is expected to have similar attributes to existing state stores with some variation:
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: filestore
+spec:
+  type: state.filestore.<providerName>
+  version: v1.0 # Following the API versioning convention proposed in https://github.com/dapr/proposals/pull/87
+  metadata:
+    # Various properties necessary for component registration
+```
+
+## Implementation Overview
+As touched on above, there are a few guiding principles I want to carry through from previous iterations of this 
+concept:
+- All operations should be asynchronously performed with the sidecar
+- When persisting or retrieving state, this should be done using an asynchronous stream to maximize SDK and runtime
+performance while minimizing resource overhead and preventing memory exhaustion
+- While SDKs should allow metadata to be stored alongside the data, if it's not supported by the provider, the 
+documentation should clearly state this lack of support (e.g. that the runtime will not otherwise persist it).
+
+### Make Runtime APIs Simpler
+I'd like to make the runtime APIs as simple as possible so they're more readily used in a composable way alongside other
+Dapr APIs by the SDKs, pluggable components or other developer-written utilities. With that in mind, I do not include
+any support for persisting metadata with the files or listing the files in the store in this API. This ensures
+broader compatibility with possible stores for file-based data and leaves the possibility of greater sophistication
+to the SDKs to handle instead.
+
+This approach allows for a leaner, more provider-agnostic API, takes work off the plate of the runtime developers and
+better positions the SDKs to provide a more comprehensive experience leveraging existing building blocks, improving
+and simplifying the developer experience.
+
+> **NOTE** It would be imperative to call out in the developer documentation for this building block that the metadata and listing
+information returned is not actually provided by the state provider and is instead provided based on values provided 
+by the developer and are not necessarily reflective of the actual state of the store. This could provide a key
+distinction between this and an actual object or blob store for those that need that level of detail (with the tradeoff
+that fewer providers would be supported in the other implementation).
+
+
+#### Example: C# SDK
+For example, in the C# SDK, the developer might specify a key/value store of their choosing to persist file metadata
+during the dependency injection registration alongside the name of the file store component:
+
+```csharp
+services.AddDapr()
+    .WithFileStore(opt => 
+    {
+        opt.FileStoreName = "my-file-store"; // Defines the file store to use for this injected `DaprFileStore` type
+        opt.MetadataStoreName = "my-metadata-store"; // Defines the metadata store to persist metadata to
+    });
+```
+
+This would provide a simple approach that allows the developer to use the SDK to persist files that optionally include
+metadata with a DI-registered type. This would avoid the current approach of having to pass in the store names each time.
+
+```csharp
+public sealed class MyClass(DaprFileStore fileStore)
+{
+    public async Task SaveDataAsync(string name, byte[] data) 
+    {
+        await fileStore.SetAsync(name, data);
+    }
+    
+    public async Task SaveDataWithMetadataAsync(string name, byte[] data, IReadOnlyDictionary<string, string> metadata )
+    {
+        await fileStore.SetAsync(name, data, metadata);
+    }
+}
+```
+
+Further, the C# SDK could provide a trivial expansion on this idea to allow named types for saving data to different 
+places throughout the application. This example uses values from a shared configuration while still relying on Dapr to 
+manage the type lifecycle:
+
+```csharp
+services.AddDapr()
+    .WithFileStore("fileStoreA", (serviceProvider, options) => 
+    {
+        // Populate the file store and metadata store names from a configuration
+        var configuration = serviceProvider.GetRequiredService<IConfiguration>();
+        var fileStoreName = configuration.GetValue<string>("FileStoreName");
+        var metadataStoreName = configuration.GetValue<string>("MetadataStoreName");
+        
+        options.FileStoreName = fileStoreName;
+        options.MetadataStoreName = metadataStoreName;        
+    })
+    .WithFileStore("fileStoreB", options => 
+    {
+        // Use static values to populate the names in this named `DaprFileStore` client
+        options.FileStoreName = "my-file-store";
+        options.MetadataStoreName = "my-metadata-store";
+    });
+```
+
+Because I don't include list support in this API either, the SDK can again be relied on to abstract this away for the 
+developer. Paired with the next-generation key/value store I [have proposed](https://github.com/dapr/dapr/issues/7338),
+the developer might query the list of files in the store by instead querying the metadata key/value store for any
+keys with the file name prefix they're interested in:
+
+```csharp
+public sealed class MyClass(DaprFileStore fileStore)
+{
+    public async Task<IReadOnlyCollection<string>> GetFileNamesAsync(string prefix)
+        => await fileStore.ListFilesAsync(prefix);
+}
+```
+
+### gRPC APIs
+In the Dapr gRPC APIs, we'd either extend the `runtime.v1.Dapr` service to add new methods or create a separate service
+specific to the file store. The benefit of taking the latter approach is that it need only be updated when the API
+changes and otherwise won't trigger downstream operations monitoring changes to that file. However, this approach
+generally falls outside the scope of this proposal and we'll assume that `runtime.v1.Dapr` is simply extended with the
+following:
+
+| Note: Consistent with current practice, APIs will have Alpha1 suffixed to the type names while in preview.
+
+| Note: Any authentication behaviors are maintained in the component YAML configuration.
+
+```proto
+// Existing Dapr service
+service Dapr {
+    rpc SetFileAlpha1(SetFileRequest) returns (SetFileResponse) {}
+    rpc GetFileAlpha1(GetFileRequest) returns (GetFileResponse) {}
+    rpc DeleteFileAlpha1(DeleteFileRequest) returns (DeleteFileResponse) {}
+}
+
+message SetFileRequest {
+    // Request details. Must be present in the first message only.
+    SetFileRequestOptions options = 1;
+    // Chunk of data of arbitrary size
+    common.v1.StreamPayload payload = 2;
+}
+
+message SetFileRequestOptions {
+    string component_name = 1 [json_name="componentName"];
+    string file_name = 2 [json_name="fileName"];
+    bool overwrite = 3 [json_name="overwrite"];
+}
+
+message SetFileResponse {}
+
+message GetFileRequest {
+    string component_name = 1 [json_name="componentName"];
+    string file_name = 2 [json_name="fileName"];
+}
+
+// The response for GetFileRequest
+message GetFileResponse {
+    // Chunk of data
+    common.v1.StreamPayload payload = 1;
+}
+
+message DeleteFileRequest {
+    string component_name = 1 [json_name="componentName"];
+    string file_name = 2 [json_name="fileName"];
+}
+
+message DeleteFileResponse {}
+```
+
+### HTTP APIs
+
+#### Persist the file data whether it already exists or not
+##### Request
+PUT `http://localhost:{daprPort}/v1.0-alpha1/state/filestore/{componentName}/{fileName}`
+
+The request body should contain an array of (nominally UTF-8 encoded) bytes.
+
+##### Response
+| Code | Description |
+| ---- | ---- |
+| 200 | File data successfully persisted |
+| 400 | File store provider not found |
+| 500 | Request formatted correctly, but error in Dapr code or underlying component | 
+
+
+#### Persist the file data only if it doesn't already exist with the given name
+##### Request
+POST `http://localhost:{daprPort}/v1.0-alpha1/state/filestore/{componentName}/{fileName}`
+
+The request body should contain an array of (nominally UTF-8 encoded) bytes.
+
+##### Response
+| Code | Description |
+| ---- | ---- |
+| 200 | File data successfully persisted |
+| 400 | File store provider not found |
+| 409 | File already exists |
+| 500 | Request formatted correctly, but error in Dapr code or underlying component |
+
+
+#### Retrieve the file data
+##### Request
+GET `http://localhost:{daprPort}/v1.0-alpha1/state/filestore/{componentName}/{fileName}`
+
+##### Response
+The response to a decryption request will have its content type header set to `application/octet-stream` as it
+returns an array of bytes representing the file data.
+
+| Code | Description |
+| ---- | ---- |
+| 200 | File data successfully retrieved |
+| 400 | File store provider not found |
+| 404 | File not found |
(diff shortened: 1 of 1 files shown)


olitomlinson:
I'm massively in support, but how does this differ from the [Object Store](https://github.com/dapr/proposals/pull/18) proposal?

WhitWaldo:
> I'm massively in support, but how does this differ from the [Object Store](https://github.com/dapr/proposals/pull/18) proposal? (Other than no support for metadata, anything else?)

There are a few differences:

1) This proposal does not anticipate ever supporting a list operation so as to be more readily and broadly supported by those providers without such capability. Most specific object and blob stores that come to mind _do_ offer such a feature. Leaving this feature as a possible differentiator for a future object/blob store API, although this would limit it to a smaller set of matching providers, is a fine trade-off to simply do without altogether here. This is looking to be little more than a provider to store large files in a way the current state store cannot and without all the other current state management add-ons.
2) This does not purport to offer those behaviors that might be more specific to object and blob stores to perform operations on data through signed URLs. Again, that might be a fine feature to use in a future state store that's more narrowly tailed to that sort of operation. This isn't that.
3) As you indicated, object and blob stores often persist and maintain a lot of metadata. In my experience, blob stores mostly just store it, but object stores will often act on it (e.g. checksum validation). No need to deal with any of that here, including several of the points brought up in your linked discussion (e.g. `Content-Length`, `Content-Hash`, `ETag`, and other metadata being used for other extraneous purposes).
4) We talked about my goal here to avoid having the SDKs deal with serialization here. An object or blob store often handles unstructured data in some format or another and I think we should absolutely create more specialized data stores that support operations more suited to one type or another (certainly could be useful from an agentic tooling and pluggable component perspective), but here, in the name of simplification and starting with a low threshold, I would like to put the responsibility on the developer for ensuring that their data can be serialized and encoded and have the API exclusively persist, retrieve and delete that data with no room for any other possibilities.
5) Object and blob store often support hierarchical or operational permissions structures such as append-only writes, write-only permissions (e.g. no deletion via API), etc. That's also intentionally excluded from consideration here.

Put more simply - those other stores anticipate the developer wanting to do both simple and far more advanced operations with their data. I'd certainly like to build more specialized data stores to accommodate such requirements, but this proposal seeks to do away with any complexities and do one thing really well: manage the reading, writing and deletion of large files in a resource-limiting and highly performant manner which is not possible in today's Dapr state management.

WhitWaldo:
> Seems like the existing s3 and other existing bindings could be mapped. Have you tried a PoC?

@lindner  The first step is proposing the shape of the block (as I've done here) and soliciting public feedback on the API shape and try to discern if anything else seems necessary within the described purpose of the API.

Out of the box, I'd certainly like to target support for Azure Blob Storage and provide an S3-compatible component (as this would facilitate connectivity with S3 itself, but also the many providers that offer S3-compatible APIs).

Next steps are getting tentative maintainer sign-off (no point building a POC if it's not going to be accepted) and then starting development of it - as I indicated in Discord, I intend to build this out as part of the next Dapr release (1.18).

olitomlinson:
In the context of its usage for storing large activity inputs and outputs in Workflows, I would strongly recommend that this design allows a workflow author to programmatically choose the path/directory to the binary file. 

This is to support multi-tenant use-cases where each tenants data MUST be stored in different locations. 

`/store/tenant-a/`

`/store/tenant-b/`

`/store/tenant-c/`

Having this location set at the time of **scheduling** the workflow (not registering the workflow) gives a good level of flexibility.


---

```c#
builder.Services.AddDaprWorkflow(options =>
    {
         options.RegisterWorkflow<MyWorkflow>( BinaryStoreName = "my-binary-store");
    }
    
    ...
    var tenantId = "tenant-a";
    var workflowId = "2c0882d7";
    
    await workflowClient.ScheduleNewWorkflowAsync(
            name: nameof(MyWorkflow),
            instanceId: workflowId,
            input: orderInfo,
            InputOutputBinaryStorePath: $"/store/{tenantId}/wf/{workflowId}"
            );
```

In the example above,  assuming we're using an S3 Binary Store, the Activity input / output blobs would be stored in the following **location**

`/store/tenant-a/wf/2c0882d7/activity/{activity-id}/output/`
`/store/tenant-a/wf/2c0882d7/activity/{activity-id}/input/`

There is an assumption that workflows have an implicit **Activity Id** which uniquely identifies each activity call. We use that Activity Id, in the path above.

---

Building on the above example, the **Reference** to the blob becomes `{app-id}||{binary-store-name}||{location}||{file-id}`

`myApp||my-binary-store||/store/tenant-a/wf/2c0882d7/activity/123/input/xyz`

The **Reference** is what is encoded in the Workflow History, rather than the blob contents.

The SDK can then dereference the data whenever the user demands it throughout the workflow. It may even be the case that the data is _never_ dereferenced, until end of the Workflow when someone requests the **output** of the completed workflow, which maybe one (or more) large blobs!

WhitWaldo:
> In the context of its usage for storing large activity inputs and outputs in Workflows, I would strongly recommend that this design allows a workflow author to programmatically choose the path/directory to the binary file.
> 
> This is to support multi-tenant use-cases where each tenants data MUST be stored in different locations.
> 
> `/store/tenant-a/`
> 
> `/store/tenant-b/`
> 
> `/store/tenant-c/`
> 
> Having this location set at the time of **scheduling** the workflow (not registering the workflow) gives a good level of flexibility.
> 
> ```cs
> builder.Services.AddDaprWorkflow(options =>
>     {
>          options.RegisterWorkflow<MyWorkflow>( BinaryStoreName = "my-binary-store");
>     }
>     
>     ...
>     var tenantId = "tenant-a";
>     var workflowId = "2c0882d7";
>     
>     await workflowClient.ScheduleNewWorkflowAsync(
>             name: nameof(MyWorkflow),
>             instanceId: workflowId,
>             input: orderInfo,
>             InputOutputBinaryStorePath: $"/store/{tenantId}/wf/{workflowId}"
>             );
> ```
> 
> In the example above, assuming we're using an S3 Binary Store, the Activity input / output blobs would be stored in the following **location**
> 
> `/store/tenant-a/wf/2c0882d7/activity/{activity-id}/output/` `/store/tenant-a/wf/2c0882d7/activity/{activity-id}/input/`
> 
> There is an assumption that workflows have an implicit **Activity Id** which uniquely identifies each activity call. We use that Activity Id, in the path above.
> 
> Building on the above example, the **Reference** to the blob becomes `{app-id}||{binary-store-name}||{location}||{file-id}`
> 
> `myApp||my-binary-store||/store/tenant-a/wf/2c0882d7/activity/123/input/xyz`
> 
> The **Reference** is what is encoded in the Workflow History, rather than the blob contents.
> 
> The SDK can then dereference the data whenever the user demands it throughout the workflow. It may even be the case that the data is _never_ dereferenced, until end of the Workflow when someone requests the **output** of the completed workflow, which maybe one (or more) large blobs!

Might this instead be done more like how actors currently stores state in KVs? Set a path on the component at registration  time that's used as the root and defer to the workflow to pick an appropriate path to save the reference to relative to the registration path? Presumably the runtime would pick a path referencing the workflow ID and any namespace values itself and then the user needn't figure out how to specify their own paths?

acroca:
First of all, I really like this proposal and I'd like to see it move forward. The API itself looks right to me: minimal, streaming, no list or metadata. My comments below are only about the workflow use case discussed in this thread, and I know I'm diverging a bit from the direction of that discussion, so sorry in advance. I hope it's still useful.

About using this store for large activity inputs and outputs, with the SDKs storing the data and putting references in the workflow history (the helper tool idea, and the `{app-id}||{store}||{location}||{file-id}` format @olitomlinson suggested): I think that specific part should be handled transparently by the runtime instead. Those references would live in workflow histories forever, and every SDK would need to understand the format written by the others. From the SDK point of view, big payloads in workflows, activities and events should just flow normally, and the runtime should be the one optimizing to not store them in the state store. (There's some early work in that direction in dapr/dapr#10170.)

That said, I think this binary store can still give workflows something very useful, close to the helper tool @WhitWaldo described, but without references going through the workflow at all: a simple API on the activity context, something like `ctx.StoreBinary(data, "myfile")` and `ctx.GetBinary("myfile")`. Activities run only once (no replay), so these are plain calls, nothing special needed in the SDKs. The workflow body doesn't know about the blobs at all: activities that need "myfile" just use the name in their code, and data is only fetched when an activity asks for it.

The sidecar would store these blobs namespaced by workflow instance, similar to what @WhitWaldo suggested above about the runtime picking paths from the workflow ID and namespace. That also means the runtime can empty the namespace when the workflow is purged, so nothing leaks. And since the name is the whole contract, workflows can cross SDKs freely.

This could also cover the scheduling case. If I want to start a workflow with a 2GB file available from the beginning, instead of sending it as the input payload I could do something like `.WithBinary("myfile", fileContents)` when scheduling it. The file gets uploaded to the workflow's namespace first, and the workflow starts once it's there.

I believe this keeps the multi-tenant requirement covered too: paths derived from namespace and instance ID keep tenant data separated, and components are namespaced anyway. If someone needs a specific blob in a different store or location, they can always call the binary store API directly from an activity.

Two small things this would need:

- A way in this API to delete all files under a prefix, so the runtime can clean up a workflow's namespace on purge. Everything else is already covered by the proposed API.
- For the workflow use, a single binary store component marked in its config, the same way the actor state store works today (`actorStateStore: true`, only one allowed). The runtime resolves it, so nothing anywhere depends on a component name.

olitomlinson:
**In the context of workflows :**

Ultimately, I really don't care which part of dapr is responsible for streaming the data into the binary store, as long as it is efficient, **and the data goes where I say it goes** :) but in the spirit of how dapr has always worked, I am slightly biased to having the runtime interfacing with the binary store target

> and the data goes where I say it goes

I can't understate this enough. 

In large Enterprise, folks want control over **where** their data goes based upon who the tenant is, and/or the sensitivity of the data.

Right now, there is **no distinction** between workflows **Control plane** data and workflows **data plane** data, but I think this is prime opportunity to make that distinction by introducing the `InputOutputBinaryStorePath` option when _scheduling_ a workflow instance, which gives a clear separation to where all inputs and outputs are stored (not in the workflow history aka the workflow control plane data)

It allows me, the author, ensure that the sensitive bits of a workflow which may contain customer IP and secrets (Inputs and Outputs), are segregated, into some logical, well-known, structure in the binary store. 

To go a step further, I would even prefer to be allowed to specify the binary store itself when scheduling a workflow for real physical segration of workflow activity inputs and outputs.

```
await workflowClient.ScheduleNewWorkflowAsync(
            name: nameof(MyWorkflow),
            instanceId: workflowId,
            input: orderInfo,
            InputOutPutBinaryStoreName: "my-binary-store",
            InputOutputBinaryStorePath: $"/store/{tenantId}/wf/{workflowId}"
            );
```

---

@acroca 

> a simple API on the activity context, something like `ctx.StoreBinary(data, "myfile")` and `ctx.GetBinary("myfile")`. Activities run only once (no replay), so these are plain calls, nothing special needed in the SDKs.

While I do agree that `Property Bag` approach this might provide value in some scenarios, this suggestion does not address the **original** problem which is how to **transparently** reference and deference large workflow Inputs and Outputs.

lindner:
Off the cuff idea: use multiple data stores, one per tenant, then have a mechanism to map that via a policy. Having magic paths feels too tied to the storage implementation. 

If needed a separate routable storage component could encapsulate multiple lower level storage components with a standard resolver mapping based on input criteria.  This may avoid the confused deputy pattern

WhitWaldo:
@acroca I think the workflow integration and how it works is probably better suited to a separate proposal, but..

I certainly agree it would be more performant if runtime entirely handled the process of identifying a large (potentially configurable) object and automatically saved to the binary store and persisted the data reference into the workflow history for precisely the reasons you said.

However, I think the SDK does need to have some visibility into the data references around how the data is retrieved back into the workflow or activities - this should ideally be loaded lazily at the developer's explicit discretion to avoid unnecessary round trips of reloading data that's constantly discarded between replays.

However, I disagree that the data should be stored by namespace because it feels like the initial use-case is now driving the shape of the parent API. Not all binary store providers support deletion by prefix, so while it might certainly be a helpful convenience, adding that to the API would likely subsequently limit the number of eligible providers. 

Rather, I don't see any reason why the runtime, when purging state, can't walk through all the data references in the workflow history and just delete each blob specifically, or maintain a list of references in key/value store automatically and track there. 

Further, as this would be a standalone building block, SDKs would be expected to offer API access to store/retrieve these files directly. Whether automatically done by the runtime when a workflow completes and returns or explicitly by the developer via the SDK, there should be a gRPC method available by the runtime (independent of this building block API) that provides for a copy operation between a workflow data reference to a specific path so workflow outputs survive history (and data ref) purges.

--

Again, doesn't really belong in this proposal, but I agree with @lindner that we should get away from the "one component for all things" mindset in the workflow configuration. It potentially hampers performance (different actors and workflows are all required to use the same storage instance across the entire Dapr implementation) and doesn't offer the flexibility I would expect end users to expect (and potentially enterprises to need) which would hamper adoption. 

In something relevant to this specific proposal though, I'm certainly not opposed to reimagining the component metadata to reflect something like "availableForWorkflow: true" as a whitelisting option by the infra team and then have the developer specify the name of this component in Workflow options when the workflow is launched. That leaves developers free to use any similarly marked binary store component in different workflows. This speaks to @olitomlinson 's concerns in that part of the binary store block configuration might be an optional "prefix" path that's used when any blobs are written, letting infrastructure dictate where the block stores data (regardless of whether it's written by the SDK directly or the runtime via Workflows).

olitomlinson:
@lindner 

I appreciate your feedback, and yes, I generally agree that magic numbers and magic strings/paths are asking for trouble, but in this case, IIRC the Binary Store component would only ever target large object stores like S3 and Azure Blob Storage, which are path-based.

But if that's not the case, then yes, I fully agree with some kind layer between!

lindner:
For prior art to look at us the Query api for state stores as a potential way to handle meta data mapping/lookup.  A similar capability model could exist for blobs and alignment between these two componen types for naming/api surface would be useful

JoshVanL (commented):
lgtm 👍

20250823-BCS-file-store.md
```diff
@@ -0,0 +1,307 @@
...
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: binarystore
+spec:
+  type: state.binarystore.<providerName>
+  version: v1.0 # Following the API versioning convention proposed in https://github.com/dapr/proposals/pull/87
+  metadata:
+    # Various properties necessary for component registration
+```
```
mikeee, line 53: Would be good to clarify whether a component can be defined with a root/key prefix, or another provider-specific keyspace, as a lightweight isolation boundary without requiring a separate bucket or container. Caller-provided keys then should be resolved within that boundary and should not be able to escape or override it.

mikeee (commented):
The proposal sounds good and the commentary offers plenty of justification into design decisions.

Just a q on component defintiions -

20250823-BCS-file-store.md
WhitWaldo: Are you referring to the type name?

mikeee: no - moreso the 'path'. At the moment it feels like the proposal refers to a bucket and a file_name with no isolation on the dapr-layer. I feel like defined components should be able to provide a path within a bucket for example that documents are stored to, with scoped access.

e.g. could I specify a `rootPrefixes` with `logs/` to scope the component to the logs folder with only the logs service able to access it rather than being a purely api concern. If the request specifies a path within the file_name it should be disallowed

WhitWaldo: I guess I more assume that infra teams wishing to impose isolation would simply spin up different instances of their storage service (accessed by different secrets) and limit access that way. While a root prefix can certainly impose a logical isolation, it's strictly applicable to the Dapr runtime and readily bypassed by just about anyone else.

I'm inclined, in the interest of not expanding requirements at the 11th hour, to decline adding such support for now. This will undoubtedly be kept in an alpha state for some time.  If there's demand, that's something that's trivially added in a future release. And if added, it should be done in a manner and with a nomenclature consistent with other building blocks.

20250823-BCS-file-store.md
```diff
@@ -0,0 +1,307 @@
...
+
+| Note: Consistent with current practice, APIs will have Alpha1 suffixed to the type names while in preview.
+
+| Note: Any authentication behaviors are maintained in the component YAML configuration.
+
+```proto
+// Existing Dapr service
+service Dapr {
+    rpc SetBinaryFileAlpha1(stream SetBinaryFileRequest) returns (SetBinaryFileResponse) {};
+    rpc GetBinaryFileAlpha1(GetBinarFileRequest) returns (stream GetBinaryFileResponse);
```
cicoyle:
```suggestion
    rpc GetBinaryFileAlpha1(GetBinaryFileRequest) returns (stream GetBinaryFileResponse);
```

20250823-BCS-file-store.md
```diff
@@ -0,0 +1,307 @@
...
+message SetBinaryFileRequest {
+    // Request details. Must be present in the first message only.
+    SetBinaryRequestOptions options = 1;
+    // Chunk of data of arbitrary size
+    common.v1.StreamPayload payload = 2;
+}
+
+message SetBinaryFileRequestOptions {
+    string component_name = 1 [json_name="componentName"];
+    string file_name = 2 [json_name="fileName"];
```
cicoyle, line 179: Are there any limitations on `fileName`? Specifically asking bc Im thinking about `fileNames` with `/` and `..` in it.

20250823-BCS-file-store.md
```diff
@@ -0,0 +1,307 @@
...
+    bool overwrite = 3 [json_name="overwrite"];
```
cicoyle, line 180: Since unset resolves to false, meaning no overwrite - I imagine we will return the `ALREADY_EXISTS` && HTTP 409 - just calling it out so its documented :)

20250823-BCS-file-store.md
```diff
@@ -0,0 +1,307 @@
...
+## Component YAML
+This component is expected to have similar attributes to existing state stores with some variation:
+
+```yaml
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: binarystore
+spec:
+  type: state.binarystore.<providerName>
```
cicoyle, line 49: Do we need to worry about etags / conditional writes here if the provider supports it? Mostly thinking about two callers overwriting the same file

20250823-BCS-file-store.md
WhitWaldo: I was thinking that's entirely all right. Services like Azure Blob Storage don't have formal folders and instead just rely on the "directories" implied by the names. I figured this left the flexibility to developers to use directories as they wanted (or didn't) and it's on the implementation of the actual provider's component to handle directories correctly (including creating them if they don't already exist upon a write).

20250823-BCS-file-store.md
WhitWaldo: Precisely

20250823-BCS-file-store.md
WhitWaldo: I wasn't planning to address it in v1. If there's demand for it (especially after a survey of the initial provider implementations and whether they support it or not), it's certainly something that can come in a later alpha version.

20250823-BCS-file-store.md
olitomlinson: I'm with Mike on this one, I think this is essential for logical isolation. 

Organisations, particularly Enterprise, need to have more granularity on how stuff is stored.

For example, if an org has an existing bucket that they want to re-use (highly likely), but mandate that all usage goes into a specific path, then this would be required and enforced at operator/infra level, rather than entrusting Devs to do it at the code-level. (Think about how Dapr creates its own schema/Table in an existing database)

@WhitWaldo @mikeee

20250823-BCS-file-store.md
cicoyle: Can you add it to a future section or somewhere so we don't forget?

20250823-BCS-file-store.md
olitomlinson: @WhitWaldo checking you saw my comment above?

WhitWaldo: @olitomlinson At this point, just trying to get the block itself added and working. You definitely represent demand for the feature, and I have no opposition to adding the feature, but in the interest of limited time, I just don't want to commit to the kitchen sink at the 11th hour in the first release. If I have the time, certainly!

20250823-BCS-file-store.md
WhitWaldo: Added a note

20250823-BCS-file-store.md
WhitWaldo: I've added a note about this to the propsoal

nelson-parente (commented):
I reviewed the implementation in dapr/components-contrib#4559 and dapr/dapr#10367 against this text at 65ea637. The RPCs, streaming directions, `Alpha1` suffixes and error codes match. The code is now ahead of the proposal in several places, and one example in the text does not work against the code.

**1. The component YAML in the proposal fails to load.** It says `type: state.binarystore.<providerName>`. daprd matches component categories by prefix and checks `state.` before `binarystore.`, so that type goes to the state store registry and fails with `couldn't find state store`. The code registers `binarystore.<provider>`, which matches `conversation.*` and `crypto.*`. Change the text to that. https://github.com/dapr/proposals/blob/65ea637cfb397067f956e1232198b956c2e27928/20250823-BCS-file-store.md#L45-L53

**2. `prefix` shipped, but the proposal lists it as a future idea, and it is not yet a boundary.** All six providers in #4559 have a `prefix` metadata field. `ObjectPath` joins it to the file name without rejecting `..`. On the flat object stores that gives an odd but contained key. Data Lake has real directories, and nothing stops `../other-tenant/x` from reaching its path API. This is the isolation Mike and Oli asked for in the open thread. Once `..` and empty segments are rejected, `prefix` becomes that boundary at almost no cost. Document it as an operator-set boundary with that rule, or take `prefix` out of the providers for alpha. https://github.com/dapr/proposals/blob/65ea637cfb397067f956e1232198b956c2e27928/20250823-BCS-file-store.md#L295-L301

**3. The limits on `fileName` are in a thread, not the text.** Cassie asked about `/` and `..`, and the answer was that providers decide. Point 2 depends on this rule, so write it into the `SetBinaryFileRequestOptions` section: `/` is a delimiter, and `..` and empty segments are rejected. https://github.com/dapr/proposals/blob/65ea637cfb397067f956e1232198b956c2e27928/20250823-BCS-file-store.md#L177-L181

**4. Proto and HTTP drift.**
- The code wraps `options` and `payload` in a `oneof`, which Josh asked for on #10367. The text also types the field `SetBinaryRequestOptions` while the message is `SetBinaryFileRequestOptions`. https://github.com/dapr/proposals/blob/65ea637cfb397067f956e1232198b956c2e27928/20250823-BCS-file-store.md#L170-L181
- Set returns 204 for PUT and POST, where the table says 200. 204 matches the state API, so update the table. https://github.com/dapr/proposals/blob/65ea637cfb397067f956e1232198b956c2e27928/20250823-BCS-file-store.md#L210-L234
- HTTP uploads over 4 MiB fail today under daprd's global body limit, which contradicts the multi-gigabyte goal. That is a runtime fix on #10367, and the text needs to state the HTTP size rule it settles on.

**5. Retries, timeouts and observability are not covered.** The runtime never retries `Set`, because the stream cannot be replayed. It rejects an upload when the first frame is more than 5 seconds late, and it wraps `Get` in the resiliency runner. None of this is in the text, and each is a contract SDK authors need. The search and vector proposal defers metrics and tracing to beta in one line. The same line works here.

Cassie's changes-requested review from Aug 11 is still the only decision on this PR. Three of her four asks are in the text. The `fileName` one is point 3. @WhitWaldo, once the text reflects points 1 to 5, @cicoyle can re-review.

WhitWaldo:
@nelson-parente To be fair, the code often supersedes the proposal as the latter is written before encountering all the ins and outs of the existing logic (especially as I've got no familiarity with the runtime core in the first place). The proposal is intended to get initial runtime maintainers approval that it's work that they would tentatively accept and isn't (at least hasn't) been a living document reflecting the current state. Rather, we have traditionally accepted the PRs as part of the post-release work and focused on getting the documentation in a good place.

1. I thought I had a unit test that validated this - I'll go back and check, but given that this is a `state` block, even if not one of the original state blocks, doesn't it make more sense to nest it (and the upcoming vector store) under `state` instead of introducing textual analogies of the stateful concepts under brand new prefixes? As in the original proposal tentatively approved by the runtime maintainers, It makes more sense to me that the runtime should be corrected to work with `state.X` and not introduce `binarystore.`, but I defer to runtime on how to proceed.
2. Yep, this wound up being an extension to the idea that seemingly everyone I spoke to about it thought would be a good idea (including . The cost of implementing it was low, so I added it.
3. That's fair that the proposal is out of sync with the code. As you noted for #2, the prefix wasn't part of the original proposal, so it lacks any information on this count.
4. Good catch - I'll refactor that.
5. Another proposal update - I can do that.

My takeaway is that there are a few consistencies to review and opportunities to really make sure the details are called out in the documentation. I'll come back and update the proposal as I've got the time.

PR #98, opened 2026-02-03
docs(proposal): multi-app protocols

tomflenner:
Create a new proposal that target to answer the need of https://github.com/dapr/dapr/issues/8897


olitomlinson:
I wanted to do this in something I'm working on.

Instead I've got to deploy 2 copies of the app with different app-ids (one for http and one for grpc). It's not the end of the world, but it does feel messy (having the protocol bleed into the app name)

tomflenner:
@olitomlinson It duplicates the ressource consumption, there is definitely a miss here IMHO

PR #101, opened 2026-03-05
WorkflowAccessPolicy

JoshVanL:
(this proposal has not been one-shot).

### DIFF
--- a/20260304-RB-workflow-access-policy.md
+++ b/20260304-RB-workflow-access-policy.md
@@ -6,7 +6,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


JoshVanL:
Thanks @olitomlinson  :) 

 I've updated the proposal to add an operation field to each operation rule, which defaults to start if omitted. We can extend to management operations like pause, resume, terminate, etc. later. Also added a note on how the intersections rules work.

  re: CRD ergonomics vs in-code policies, I think that you probably want both. It is the case that the infrastructure person does want to manage RBAC across a namespace, see the MCP server propsoal as one example https://github.com/dapr/proposals/pull/100

cicoyle (changes requested):
Few things from me, I like the general idea, but think we could still tweak the design to offer a unified approach for access policy in Dapr. lmk what you think!

famarting (approved):

famarting:
qq, would it be desirable to use this access policy resource as another layer of authorization for https://github.com/dapr/proposals/blob/main/20260304-CR-mcpserver.md , and if so, would it make sense to overlay an abstraction on top of the WorkflowAccessPolicy resource to make authorization of mcpservers a first class citizen?

PR #104, opened 2026-03-18
Proposal: Workflow external event timer tracking

acroca:
Today there is no reliable way to tell from the outside which workflow instances are waiting for an external event, or which event they are waiting for. Some SDKs set the timer name to the event name, but that is not enough to distinguish a `waitForExternalEvent` timer from a regular timer. On top of that, SDKs that allow indefinite waits (no timeout) don't create a timer at all, making those waits completely invisible.

This proposal addresses both gaps by adding a structured `timer_context` field to `CreateTimerAction` and `TimerCreatedEvent` in durabletask-protobuf, and requiring SDKs to always create a timer when `waitForExternalEvent` is called, using a far-future timestamp when no timeout is specified.

### DIFF
--- a/20260318-RS-workflow-external-event-timer-tracking.md
+++ b/20260318-RS-workflow-external-event-timer-tracking.md
@@ -1,0 +1,139 @@
+# Workflow External Event Timer Tracking
+
+Author: Albert Callarisa @acroca
+Status: Proposed
+Introduced: 2026-03-18
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


20260318-RS-workflow-external-event-timer-tracking.md
```diff
@@ -0,0 +1,139 @@
...
+Two messages in `durabletask-protobuf` are extended with a `timer_context` oneof:
+
+```protobuf
+// orchestrator_actions.proto
+message CreateTimerAction {
+    google.protobuf.Timestamp fireAt = 1;
+    optional string name = 2;
+
+    oneof timer_context {
+        ExternalEventTimerInfo external_event = 3;
```
JoshVanL:
```suggestion
        TimerOriginExternalEvent external_event = 3;
```

JoshVanL (changes requested):

JoshVanL (approved):

PR #107, opened 2026-04-28
Search & Vector Stores

mikeee:
Purely a building block implementation of the search and vector stores

### DIFF
--- a/20260428-BCRS-Search-and-vector-blocks.md
+++ b/20260428-BCRS-Search-and-vector-blocks.md
@@ -1,0 +1,221 @@
+# Search and Vector Building Blocks
+
+Author: Mike Nguyen - @mikeee (hey@mike.ee)
+
+## Overview
+
+This is a proposal surrounding the implementation of two new specialised state stores without bringing into
+scope the deep integrations planned on implementation of these building blocks. Whilst the two building 
+blocks share almost the same envelope, the method of requesting the records can vary significantly and should
+be kept distinct shapes.
+
+## Background
+
+The current implementations of state stores do not accomodate the necessity to access data not only by 
+key-value but an extended querying layer. New 'specialised' building blocks facilitate rapid development.
+
+There is a specific need from AI workloads to not only query by key but also by 'relevant' records returned
+through lexical/structured matching or geometric proximity for example.
+
+## Providers in scope
+
+| Provider | Lexical Search | Vector Search |
+|----------|---|---|
+| Meilisearch | ✓ | ✓ |
+| Elasticsearch | ✓ | ✓ |
+| Chroma | ✓ | ✓ |
+
+Other providers with existing component implementations may be leveraged.
+
+## API / Protos
+
+### Search
+
+```protobuf
+rpc IndexDocumentsAlpha1(IndexDocumentsRequestAlpha1) returns (IndexDocumentsResponseAlpha1) {}
+rpc DeleteDocumentsAlpha1(DeleteDocumentsRequestAlpha1) returns (google.protobuf.Empty) {}
+rpc SearchAlpha1(SearchRequestAlpha1) returns (SearchResponseAlpha1) {}
+
+message SearchDocument {
+  string id = 1;
+  bytes content = 2;
+  map<string, string> metadata = 3;
+}
+
+message SearchHit {
+  SearchDocument document = 1;
+  double score = 2;
+  map<string, string> highlights = 3;
+}
+
+enum SortOrder {
+  SORT_ORDER_UNSPECIFIED = 0;
+  SORT_ORDER_ASC = 1;
+  SORT_ORDER_DESC = 2;
+}
+
+message SortClause {
+  string field = 1;
+  SortOrder order = 2;
+}
+
+enum IndexAck {
+  INDEX_ACK_UNSPECIFIED = 0;
+  INDEX_ACK_QUEUED = 1;
+  INDEX_ACK_DURABLE = 2;
+}
+
+message IndexDocumentsRequestAlpha1 {
+  string store_name = 1;
+  string index = 2;
+  repeated SearchDocument documents = 3;
+  map<string, string> metadata = 10;
+}
+
+message IndexDocumentsResponseAlpha1 {
+  repeated string failed_ids = 1;
+  IndexAck ack = 2;
+}
+
+message DeleteDocumentsRequestAlpha1 {
+  string store_name = 1;
+  string index = 2;
+  repeated string ids = 3;
+  map<string, string> metadata = 10;
+}
+
+message SearchRequestAlpha1 {
+  string store_name = 1;
+  string index = 2;
+
+  oneof query {
+    string text = 3;
+    google.protobuf.Struct native = 4;
+  }
+
+  google.protobuf.Struct filter = 5;
+  uint32 top_k = 6;
+  uint32 offset = 7;
+  repeated string return_fields = 8;
+  bool include_content = 9;
+  map<string, string> metadata = 10;
+  repeated string search_fields = 11;
+  repeated SortClause sort = 12;
+  repeated string highlight_fields = 13;
+
+  reserved 14, 15, 16;
+}
+
+message SearchResponseAlpha1 {
+  repeated SearchHit hits = 1;
+  uint64 total_hits = 2;
+  string continuation_token = 3;
+}
+```
+
+### Vector
+
+```protobuf
+rpc UpsertVectorsAlpha1(UpsertVectorsRequestAlpha1) returns (UpsertVectorsResponseAlpha1) {}
+rpc DeleteVectorsAlpha1(DeleteVectorsRequestAlpha1) returns (google.protobuf.Empty) {}
+rpc GetVectorsAlpha1(GetVectorsRequestAlpha1) returns (GetVectorsResponseAlpha1) {}
+rpc QueryVectorsAlpha1(QueryVectorsRequestAlpha1) returns (QueryVectorsResponseAlpha1) {}
+rpc BatchQueryVectorsAlpha1(BatchQueryVectorsRequestAlpha1) returns (BatchQueryVectorsResponseAlpha1) {}
+
+enum DistanceMetric {
+  DISTANCE_METRIC_UNSPECIFIED = 0;
+  DISTANCE_METRIC_COSINE = 1;
+  DISTANCE_METRIC_DOT_PRODUCT = 2;
+  DISTANCE_METRIC_EUCLIDEAN = 3;
+}
+
+message SparseVector {
+  repeated uint32 indices = 1;
+  repeated float values = 2;
+}
+
+message NamedVector {
+  repeated float values = 1;
+  SparseVector sparse_values = 2;
+}
+
+message VectorRecord {
+  string id = 1;
+  repeated float values = 2;
+  bytes payload = 3;
+  map<string, string> metadata = 4;
+  SparseVector sparse_values = 5;
+  map<string, NamedVector> named_vectors = 6;
+}
+
+message VectorMatch {
+  VectorRecord record = 1;
+  double score = 2;
+}
+
+message UpsertVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated VectorRecord records = 3;
+  map<string, string> metadata = 10;
+}
+
+message UpsertVectorsResponseAlpha1 {
+  repeated string failed_ids = 1;
+}
+
+message DeleteVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated string ids = 3;
+  map<string, string> metadata = 10;
+}
+
+message GetVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated string ids = 3;
+  bool include_values = 4;
+  map<string, string> metadata = 10;
+}
+
+message GetVectorsResponseAlpha1 {
+  repeated VectorRecord records = 1;
+}
+
+message QueryVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+
+  oneof query {
+    VectorRecord vector = 3;
+    string by_id = 4;
+  }
+
+  uint32 top_k = 5;
+  google.protobuf.Struct filter = 6;
+  bool include_values = 7;
+  bool include_payload = 8;
+  DistanceMetric metric = 9;
+  map<string, string> metadata = 10;
+  SparseVector sparse_query = 11;
+  optional float alpha = 12;
+  optional string vector_name = 13;
+  optional float score_threshold = 14;
+}
+
+message QueryVectorsResponseAlpha1 {
+  repeated VectorMatch matches = 1;
+}
+
+message BatchQueryVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated QueryVectorsRequestAlpha1 queries = 3;
+  map<string, string> metadata = 10;
+}
+
+message BatchQueryVectorsResponseAlpha1 {
+  repeated QueryVectorsResponseAlpha1 results = 1;
+}
+```
--- a/20260428-BCRS-Search-and-vector-blocks.md
+++ b/20260428-BCRS-Search-and-vector-blocks.md
@@ -27,7 +27,9 @@ through lexical/structured matching or geometric proximity for example.
 
 Other providers with existing component implementations may be leveraged.
 
-## API / Protos
+## HTTP API / Protos
+
+A HTTP api will be exposed mirroring the below.
 
 ### Search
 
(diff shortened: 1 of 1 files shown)


20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
+# Search and Vector Building Blocks
+
+Author: Mike Nguyen - @mikeee (hey@mike.ee)
+
+## Overview
+
+This is a proposal surrounding the implementation of two new specialised state stores without bringing into
+scope the deep integrations planned on implementation of these building blocks. Whilst the two building 
+blocks share almost the same envelope, the method of requesting the records can vary significantly and should
```
acroca, line 9: I'm wondering if it'd make sense to ship a single building block with both of the APIs combined into one. With a single component, we get a unified API for index/upsert/delete documents, and different endpoints for fetching, one for lexical search and one for vector search.
This would be beneficial if a user needs to use both APIs at the same time.

WhitWaldo: Unless there's some scenario you can think of where a query to a provider might do some sort of join operation between lexical and vector searches (e.g. we can offload that heavy lifting to the provider - we absolutely should _not_ do any manipulation of this data in the runtime), I would vote they remain separate.

This obviates the need for another (and differing) capabilities functionality at the building block level, it keeps the APIs for each building block simple and focused and it potentially improves the rate at which provider implementations can be developed and added (e.g. they need only implement the API associated with one block at a time, not have to implement a broader set just for v1.0).

Especially if the results from either are instead used alongside local logic in the app, that's the domain of the SDK to offer a usage smooth experience where it's the domain of the runtime to establish connectivity, issue commands and perform data transfer. I think runtime's role is better suited to small, lean building block APIs going forward.

mikeee: I was considering this too - I can make it clearer in the proposal but I share the same sentiment as Whit. 

The two building blocks can be objectively very different. I feel like it may be running the risk of becoming a feature flag mess/very confusing for implementation where providers have differing feature sets that fall distinctly within the lexical or vector search buckets. 

We should focus on having distinct building blocks, if you need to use both APIs at the same time then you'd just need to register another component type with the same provider.

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+  INDEX_ACK_UNSPECIFIED = 0;
+  INDEX_ACK_QUEUED = 1;
+  INDEX_ACK_DURABLE = 2;
+}
+
+message IndexDocumentsRequestAlpha1 {
+  string store_name = 1;
+  string index = 2;
+  repeated SearchDocument documents = 3;
+  map<string, string> metadata = 10;
```
cicoyle:
```suggestion
  map<string, string> metadata = 4;
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+message IndexDocumentsResponseAlpha1 {
+  repeated string failed_ids = 1;
+  IndexAck ack = 2;
+}
+
+message DeleteDocumentsRequestAlpha1 {
+  string store_name = 1;
+  string index = 2;
+  repeated string ids = 3;
+  map<string, string> metadata = 10;
```
cicoyle:
```suggestion
  map<string, string> metadata = 4;
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+message VectorMatch {
+  VectorRecord record = 1;
+  double score = 2;
+}
+
+message UpsertVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated VectorRecord records = 3;
+  map<string, string> metadata = 10;
```
cicoyle:
```suggestion
  map<string, string> metadata = 4;
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+
+message UpsertVectorsResponseAlpha1 {
+  repeated string failed_ids = 1;
+}
+
+message DeleteVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated string ids = 3;
+  map<string, string> metadata = 10;
```
cicoyle:
```suggestion
  map<string, string> metadata = 4;
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+}
+
+message GetVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated string ids = 3;
+  bool include_values = 4;
+  map<string, string> metadata = 10;
```
cicoyle:
```suggestion
  map<string, string> metadata = 4;
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+
+message QueryVectorsResponseAlpha1 {
+  repeated VectorMatch matches = 1;
+}
+
+message BatchQueryVectorsRequestAlpha1 {
+  string store_name = 1;
+  string collection = 2;
+  repeated QueryVectorsRequestAlpha1 queries = 3;
+  map<string, string> metadata = 10;
```
cicoyle:
```suggestion
  map<string, string> metadata = 4;
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+
+| Provider | Lexical Search | Vector Search |
+|----------|---|---|
+| Meilisearch | ✓ | ✓ |
+| Elasticsearch | ✓ | ✓ |
+| Chroma | ✓ | ✓ |
+
+Other providers with existing component implementations may be leveraged.
+
+## HTTP API / Protos
```
cicoyle, line 70: Can you add HTTP to the proposal? I only see the grpc protos

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+  repeated float values = 2;
+  bytes payload = 3;
+  map<string, string> metadata = 4;
+  SparseVector sparse_values = 5;
+  map<string, NamedVector> named_vectors = 6;
+}
+
+message VectorMatch {
+  VectorRecord record = 1;
+  double score = 2;
```
cicoyle, line 500: Can we notate what a higher/lower score indicates for clarity?

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+  uint32 top_k = 5;
+  google.protobuf.Struct filter = 6;
+  bool include_values = 7;
+  bool include_payload = 8;
+  DistanceMetric metric = 9;
+  map<string, string> metadata = 10;
+  SparseVector sparse_query = 11;
+  optional float alpha = 12;
+  optional string vector_name = 13;
+  optional float score_threshold = 14;
```
cicoyle: Can you notate what a higher/lower score indicates for clarity?

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+
+message IndexDocumentsRequestAlpha1 {
+  string store_name = 1;
+  string index = 2;
+  repeated SearchDocument documents = 3;
+  map<string, string> metadata = 10;
+}
+
+message IndexDocumentsResponseAlpha1 {
+  repeated string failed_ids = 1;
```
cicoyle: Should this also return a reason or code so we can determine if we should retry or not? It might make sense to add something like the following (not exact syntax, but for ex):
```
repeated FailedItem {
string id; 
uint32 code; 
string message;
}
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+  repeated SortClause sort = 12;
+  repeated string highlight_fields = 13;
+
+  reserved 14, 15, 16;
+}
+
+message SearchResponseAlpha1 {
+  repeated SearchHit hits = 1;
+  uint64 total_hits = 2;
+  string continuation_token = 3;
```
cicoyle, line 303: I see offset in the req, are we intending to do offset based or cursor based pagination?

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+  optional string vector_name = 13;
+  optional float score_threshold = 14;
+}
+
+message QueryVectorsResponseAlpha1 {
+  repeated VectorMatch matches = 1;
+}
+
+message BatchQueryVectorsRequestAlpha1 {
+  string store_name = 1;
```
cicoyle, line 572: I see `store_name` && `collection` here and inside `QueryVectorsRequestAlpha1` - shall we trim the inner msg?

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+## Overview
+
+This is a proposal surrounding the implementation of two new specialised state stores without bringing into
+scope the deep integrations planned on implementation of these building blocks. Whilst the two building 
+blocks share almost the same envelope, the method of requesting the records can vary significantly and should
+be kept distinct shapes.
+
+## Background
+
+The current implementations of state stores do not accomodate the necessity to access data not only by
```
cicoyle, line 14:
```suggestion
The current implementations of state stores do not accommodate the necessity to access data not only by 
```

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+scope the deep integrations planned on implementation of these building blocks. Whilst the two building 
+blocks share almost the same envelope, the method of requesting the records can vary significantly and should
+be kept distinct shapes.
+
+## Background
+
+The current implementations of state stores do not accomodate the necessity to access data not only by 
+key-value but an extended querying layer. New 'specialised' building blocks facilitate rapid development.
+
+There is a specific need from AI workloads to not only query by key but also by 'relevant' records returned
```
cicoyle, line 17: Can you add a `Use Cases` section? I believe it would help ground the separate vs combined API decision - might be good to note under an `Alternatives Considered` section or similar. There are RAG/agent retrieval patterns where lexical & vector results get fused together. I believe its a first class AI usecase that should be considered, rather than it being an edge case. Its worth deciding if we should have the APIs separate and/or have a hybrid, even if the hybrid API is out of scope for now - it should be thought thru ahead of implementation to prevent breaking changes later on.

20260428-BCRS-Search-and-vector-blocks.md
cicoyle: One thing I'd love to see explicitly spelled out is the story for someone who wants to use both search & vector against the same provider. From ^ it sounds like the answer is to register 2 components pointing to the same server. 

When that happens, is the underlying data shared - so an item indexed for search and its vector can be tied together by a common ID or are they independent? I think that matters with AI use cases where you might want to store something once and both search it by keyword && find it by similarity.

And is there a scenario where a single query should do both at once - keyword & similarity combined by the provider itself?

20260428-BCRS-Search-and-vector-blocks.md
```diff
@@ -0,0 +1,260 @@
...
+Other providers with existing component implementations may be leveraged.
+
+## HTTP API / Protos
+
+A HTTP api will be exposed mirroring the below.
+
+### Search
+
+```protobuf
+rpc IndexDocumentsAlpha1(IndexDocumentsRequestAlpha1) returns (IndexDocumentsResponseAlpha1) {}
```
cicoyle, line 116: Should we also consider additional rpcs for `CreateIndex`, `CreateCollection`, `DescribeIndex`, `ListIndexes`, `DeleteIndexes`, etc?

cicoyle (changes requested):

20260428-BCRS-Search-and-vector-blocks.md
WhitWaldo: Rather than trying to design a feature like @cicoyle is describing as a one-off feature for this, I'd much rather take that idea back to the drawing board and figure out if the idea is applicable across more components and design that as a separate runtime feature than necessarily jam in something that potentially complex as a one-off here. Seems like that would dramatically increase the complexity of what's otherwise a pretty straightforward proposal here.

20260428-BCRS-Search-and-vector-blocks.md
WhitWaldo: One of the key ideas of my proposal to split out into discrete building blocks is to explicitly pivot away from what the key/value state store turned into: a far more use-case governed platform with runtime shims holding together what providers failed to provide.

I would dramatically prefer that if there's an AI use-case, it be proposed as a separate block - more of a composite component that's able to express the individual functionality but which only applies to those providers that can do both operations. This is a central tenant of my handily rejected [planes](https://github.com/dapr/proposals/pull/103) proposal, but is exactly along the lines of what I was thinking.

Some people will just need a vector store. Some people will just need a search store. For those that need both, build something that specifically speaks to that use-case but atop far more generic basic components.

20260428-BCRS-Search-and-vector-blocks.md
mikeee: It’s the raw value for the specified distance metric, so whether a higher or lower score is better depends on that metric—for example, lower is better for Euclidean distance.

mikeee: (I'm not planning on doing any normalisation here to reduce confusion)

20260428-BCRS-Search-and-vector-blocks.md
mikeee: I'll be updating this to cursor-based pagination with an opaque token to abstract the provider specific behaviour.

20260428-BCRS-Search-and-vector-blocks.md
mikeee: Yep, sorry I forgot to pull these methods from the draft PR

20260428-BCRS-Search-and-vector-blocks.md
mikeee: This was a design decision to keep the metadata field consistent across the new building block interface, let me know if you have any suggestions on this

20260428-BCRS-Search-and-vector-blocks.md
mikeee: Addressed in the scope boundary

nelson-parente (commented):
I reviewed the implementation in dapr/components-contrib#4613 and dapr/dapr#9974 against this text at 4339cda. The code follows the proposal closely, and the protos and HTTP routes match it one for one. The open threads are further along than the PR page shows.

**Five threads are answered in the text but still open.** @mikeee, resolve them with a pointer to the line, then @cicoyle can re-review against a short list.
- Score direction: the `VectorMatch.score` comment and the vector score table. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L709-L723
- A reason code on failed items: `FailedItem.error` is a `google.rpc.Status`. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L217-L223
- Offset or cursor: opaque continuation tokens. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L401-L426
- Inner `store_name` and `collection` on batch queries: the text says they are ignored. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L730-L731
- One block or two: the opening paragraph keeps them separate. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L7-L10

**What the text still needs.** Each item below is something the code already decided and the proposal does not say.

1. **Use Cases, and whether one index can back both blocks.** Cassie's follow-up on the one-or-two thread asked whether the two components can share data. The code answers it by accident: point the vector component at an index the search component created and `GetCollection` reports `Dimensions: 0` as if it were a valid collection. Hybrid retrieval is the main RAG pattern, and today the text only lists it as deferred. A short Use Cases section settles both threads. It needs to say what a user who wants lexical and vector results over the same documents does in alpha.
2. **An empty search query.** The `oneof` allows neither `text` nor `native`. The runtime forwards that to the component, Meilisearch treats it as match-all, and its conformance suite depends on that. The contrib doc comment says "exactly one", and the runtime validation list covers the vector `oneof` but not this one. Write the rule down. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L739-L749
3. **Inner batch fields.** Ignoring them means a caller who sets a different collection on one query gets results from another collection without an error. Alpha is the time to remove the two fields, which is what Cassie asked. The other option is to reject a value that differs from the batch.
4. **The HTTP error body.** The runtime renders a component status error with the `google.rpc.Code` name as `errorCode`, for example `NOT_FOUND`, next to the existing `ERR_*` codes. The component error contract section covers gRPC only. This is a new HTTP contract, so it belongs here. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L758-L768
5. **Token binding.** The text binds a continuation token to "component-declared result-affecting metadata", but nothing defines how a component declares it, and the Meilisearch fingerprint leaves metadata out. Define the mechanism or drop the clause. https://github.com/dapr/proposals/blob/4339cda202aa2fa80be183263cc574eeebafbd57/20260428-BCRS-Search-and-vector-blocks.md#L412-L413

Observability is deferred to beta, which is fine, and the code matches that. The code-level findings are on #4613 and #9974.

PR #108, opened 2026-06-19
Remove `model` from the Conversation building block and promote to a runtime-provided argument

WhitWaldo:
Please see the proposal for details

### DIFF
--- a/20260619-BCR-conversation-models.md
+++ b/20260619-BCR-conversation-models.md
@@ -1,0 +1,153 @@
+﻿# Remove `model` from the Conversation building block specification
+
+- Author: Whit Waldo @whitwaldo
+- State: Proposed
+- Introduced: 2026-06-19
+
+## Overview
+The Conversation building block is in an Alpha state and subject to breaking changes; this is the time to make one. When the infrastructure team defines a Conversation component today, they provide the component name, block type/version/provider, connection details (endpoint, auth), the provider configuration, Dapr functionality configuration, and the model name.
+
+I propose removing this last item. The `model` property leaves the component spec and becomes a required property on the conversation request. This is one conceptual change with the follow-through it implies: proto, runtime (including caching), SDKs, and documentation. 
+
+Downstream enhancements are deferred to separate proposals.
+
+## Background
+
+### Motivation
+The Conversation building block landed with the model named in the component, which was a sensible choice when the supported providers each exposed only a handful of models. As the provider set has grown, the most common request I've seen (both in Discord and in person) has been a way to choose the model without standing up a near-duplicate component for each one. That pressure is only increasing: Ollama already supports 100+ models and a provider like OpenRouter (120+) would make a component-per-model approach impractical for an infrastructure team to reasonably maintain.
+
+One natural response is to let the request override the component's model. I don't think this is the right direction. Overrides raise their own questions about which properties are overridable, who owns a connection whose values can change at runtime, and how Dapr functionality layered on top (caching, for example) behaves when they do. Rather than answer those, I think the cleaner move is to reconsider where the model belongs in the first place.
+
+### Model is a runtime concern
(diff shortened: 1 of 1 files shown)


acroca:
I agree with this change, the model belongs on the request.

My first reaction was that this changes the meaning of the Conversation component: without the model, it stops being "a conversation configuration" and becomes more of an "LLM provider" component (endpoint + credentials). But looking at the current implementation, that shift already happened, temperature, tools, response format, etc. are all on the request already. The model was the last piece left in the component, so this proposal is just finishing what was already started.

One thing I'd change: I think the allowlist mentioned in Future Work should ship with this proposal, not later. Today the component pins the model, so an admin controls exactly what gets used. With this change, any caller can pick any model the API key can reach, including very expensive ones. An optional allowlist on the component (default: all models allowed) would keep that control available without getting in anyone's way. Since it's optional it doesn't add friction for simple setups, but it lets component admins keep costs under control from day one instead of waiting for a follow-up proposal.

WhitWaldo:
@acroca I certainly think it would be valuable to ship something like the allowlist, but I know work was done on the allow/deny list concept for workflows this last release and I don't know where any of that landed, so I've sort of left that as an open-ended thing for runtime to tackle.

But I'd certainly like to see a proposal for it submitted alongside this to have both sets of features shipped in the same release. It doesn't make a lot of sense to rewrite the wheel if there's a suitable one already that can be slotted in here, but I just don't know enough about what already exists to speak to that or not.

mikeee:
I'm a +1 on the proposal as it currently is, agreeing strongly with Albert that we bring governance into the initial scope. 

https://github.com/dapr/dapr/blob/master/pkg/acl/workflow/policy.go

This is what landed on the last release Whit ^

20260619-BCR-conversation-models.md
```diff
@@ -0,0 +1,153 @@
...
+                }
+            }
+        )
+    ],
+    new ConversationOptions("conversation", "kimi-k2.7-code")
+);
+```
+
+### Interaction with `context_id`
+With model now per-request, a follow-up request that reuses an existing `context_id` may carry a different model than the request that opened the context. Whether switching models within a single context is meaningful is provider-defined; the runtime makes no guarantee here, consistent with the Crypto stance above. The model binds per request.
```
sicoyle, line 133: pls double check but context_id is not used anywhere IIRC

20260619-BCR-conversation-models.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Backwards Compatibility
+This is a breaking change, but the Conversation block is in alpha, so this is the time to do this.
+
+## Acceptance Criteria
```
sicoyle, line 138: dapr agents and every other agent framework leveraging this api will need to expose a new parameter upon agent instantiation to support this

sicoyle, line 138: contrib will need a bunch of updates to rm defaulting and removing this field and properly propagating this new req field over

sicoyle (commented):
few comments so far from me pls, but overall am aligned with this change. It aligns with us using openai's api for inspiration too:
https://github.com/openai/openai-go/blob/main/chatcompletion.go#L192

20260619-BCR-conversation-models.md
WhitWaldo: From a prototype level, I think it's just a refactoring to use the model from the request instead of the configuration component, except with regards to the [DeepSeek component](https://docs.dapr.io/reference/components-reference/supported-conversation/deepseek/) which doesn't expose a model at all (though their [API allows more](https://api-docs.deepseek.com/) than one since the component was authored). But DeepSeek needs updating this month anyway since it [defaults](https://github.com/dapr/components-contrib/blob/6309477028528c6027df1ab04baa2e04ace1bcdb/conversation/deepseek/deepseek.go#L39) to "deepseek-chat" and that's being [deprecated](https://api-docs.deepseek.com/) in their API later this month (7/24/26).

20260619-BCR-conversation-models.md
WhitWaldo: This might be a bleed-over from the MAF integration. I'll follow up.

